PR #3, opened 2015-04-14
Sprint 20 TLT-1172 Mailing List Refactoring

douglashall:
@rascalking 
@sapnamysore 

Refactoring this a bit to support mailing lists per section and UI enhancements.


mailing_list/management/commands/sync_listserv.py
```diff
@@ -18,20 +19,18 @@ def __init__(self, *args, **kwargs):
...
+                    ml.sync_listserv_membership()
             else:
                 for ml in MailingList.objects.all():
                     canvas_course_ids.append(ml.canvas_course_id)
-                    mailing_lists.append(ml)
+                    ml.sync_listserv_membership()
 
-            logger.info("Synchronizing listserv mailing lists for canvas course ids %s", canvas_course_ids)
-
-            sync_mailing_lists_with_listserv(mailing_lists)
```
rascalking: We're now syncing one at a time...do we care about having one of N lists synchronized, but not the rest?  I'm guessing we're just going to rerun the sync regardless, just wanted to make sure.

douglashall: Not sure I'm following you. Are you thinking of an error condition where this command would fail to run to completion?

mailing_list/listserv_clients/mailgun.py
```diff
@@ -29,11 +30,26 @@ def api_user(self):
...
+    def get_list(self, mailing_list):
+        address = mailing_list.address
+        api_url = "%s/%s" % (self.api_url, address)
+
+        with ApiRequestTimer(logger, 'GET', api_url) as timer:
+            response = requests.get(api_url, auth=(self.api_user, self.api_key))
+            timer.status_code = response.status_code
+
+        if response.status_code != 200:
+            if response.status_code != 404:
```
rascalking: why not just make this `response.status_code not in (200, 404)`?

mailing_list/management/commands/sync_listserv.py
rascalking: I just meant that previously, we found the set of lists to sync, then synced them in one call.  Now we're syncing them as we find them.  I was wondering if we cared whether we hit an error at list 2 of 4.  I don't think we do, just wanted to clarify.

mailing_list/listserv_clients/mailgun.py
douglashall: Heh, wrote that one late last night. :)

mailing_list/management/commands/sync_listserv.py
douglashall: Got it. Yes, I think we would just re-run the job. Before we were also syncing one at a time, it was just happening inside that sync_mailing_lists_with_listserv utility that I jettisoned.

mailing_list/models.py
```diff
@@ -1,33 +1,152 @@
...
+            defaults = kwargs.get('defaults', {})
+            for s in canvas_sections:
+                section_id = s['id']
+                mailing_list = mailing_lists_by_section_id.get(section_id)
+                if not mailing_list:
+                    create_kwargs = {
+                        'canvas_course_id': canvas_course_id,
+                        'section_id': section_id
+                    }
+                    create_kwargs.update(defaults)
```
rascalking: This is overwriting what's in create_kwargs with what's in defaults, which isn't how I think of defaults working.  If that's the intent, maybe rename defaults to overrides?

douglashall: Sure.

mailing_list/models.py
```diff
@@ -1,33 +1,152 @@
...
+            logger.exception("Failed to get canvas enrollments for canvas_course_id %s", canvas_course_id)
+            raise
+
+        univ_ids = []
+        for enrollment in canvas_enrollments:
+            try:
+                univ_ids.append(enrollment['user']['sis_user_id'])
+            except KeyError:
+                logger.debug("Found canvas enrollment with missing sis_user_id %s", json.dumps(enrollment, indent=4))
+        enrolled_emails = set([p.email_address for p in Person.objects.filter(univ_id__in=univ_ids)])
```
rascalking: nitpick: set comprehensions are more readable, imo

(btw, in case it's not clear, feel free to ignore all my nitpicks)

mailing_list/models.py
```diff
@@ -1,33 +1,152 @@
...
+
+        mailing_list_emails = enrolled_emails - unsubscribed
+
+        # Only add subscribers to the listserv if:
+        # 1. The subscriber is on the whitelist
+        # OR
+        # 2. Settings tell us to ignore the whitelist
+        if not hasattr(settings, 'IGNORE_WHITELIST'):
```
rascalking: so if I see `IGNORE_WHITELIST = True` in a settings file, and change it to `False`, it'll still get ignored?

mailing_list/models.py
douglashall: Nope, this is great. Thought I had caught all those already.

mailing_list/models.py
douglashall: Good catch.

mailing_list/templates/mailing_list/index.html
```diff
@@ -44,5 +21,76 @@
...
+                            <img src="{% static 'mailing_list/images/ajax-loader-small.gif' %}" ng-show="list.isUpdating"/>
+                            <span class="text-success feedback" ng-show="list.updated">&#10003;</span>
+                        </td>
+                    </tr>
+                </tbody>
+            </table>
+        </div>
+        <div id="other-sections" ng-show="ml.hasOtherSections()">
+            <h2>Other Sections</h2>
+            <table class="table">
```
rascalking: suggestion for next time: put the table generation in a separate template, then [`include`](https://docs.djangoproject.com/en/1.8/ref/templates/builtins/#include) it for each set of lists

rascalking:
:+1:

mailing_list/templates/mailing_list/index.html
douglashall: Yeah, there's a way to use angular to break this into a separate template as well. I punted for now.

PR #4, opened 2015-04-15
Sprint 20 TLT-1213 Unit Tests

douglashall:
@rascalking 
@elliottyates

### DIFF
--- a/lti_emailer/settings/unit_test.py
+++ b/lti_emailer/settings/unit_test.py
@@ -19,23 +19,3 @@ CACHES = {
         'BACKEND': 'django.core.cache.backends.dummy.DummyCache'
     },
 }
-
-"""
-Unit tests should be able to run without talking to external services like
-redis/elasticache.  Let's disable the caching middleware, and use the db
-to store session data.
-"""
-MIDDLEWARE_CLASSES = list(MIDDLEWARE_CLASSES)
-MIDDLEWARE_CLASSES[MIDDLEWARE_CLASSES.index('cached_auth.Middleware')] = \
-    'django.contrib.auth.middleware.AuthenticationMiddleware'
-MIDDLEWARE_CLASSES = tuple(MIDDLEWARE_CLASSES)
-SESSION_ENGINE = 'django.contrib.sessions.backends.db'
-
-"""
-Don't assume we'll have oauth credentials set for us in secure.py, set
-them up here instead.
-"""
-SECURE_SETTINGS.setdefault('lti_oauth_credentials', {}).update({'foo': 'bar'})
-if not LTI_OAUTH_CREDENTIALS:
-    LTI_OAUTH_CREDENTIALS = {}
-LTI_OAUTH_CREDENTIALS.update({'foo': 'bar'})
--- a/lti_emailer/settings/unit_test.py
+++ b/lti_emailer/settings/unit_test.py
@@ -14,6 +14,12 @@ DATABASES = {
     },
 }
 
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.dummy.DummyCache'
+    },
+}
+
 """
 Unit tests should be able to run without talking to external services like
 redis/elasticache.  Let's disable the caching middleware, and use the db
--- a/mailing_list/listserv_clients/tests.py
+++ b/mailing_list/listserv_clients/tests.py
@@ -1,0 +1,282 @@
+import json
+
+from django.test import TestCase
+
+from mock import patch
+
+from icommons_common.utils import Bunch
+
+from .exceptions import ListservApiError
+from .mailgun import MailgunClient as ListservClient
+from mailing_list.models import MailingList
+
+
+class ListservClientTests(TestCase):
+    fixtures = [
+        'mailing_list/fixtures/test_mailing_list.json'
+    ]
+    longMessage = True
+
+    @patch('requests.get')
+    def test_get_list_success(self, mock_requests_get):
+        listserv_client = ListservClient()
+        response = Bunch(
+            status_code=200,
+            json=lambda: {'list': {}}
+        )
+        mock_requests_get.return_value = response
+        mailing_list = MailingList.objects.get(id=1)
+        api_url = "%s/%s" % (listserv_client.api_url, mailing_list.address)
+        auth = (listserv_client.api_user, listserv_client.api_key)
+
+        result = listserv_client.get_list(mailing_list)
+
+        mock_requests_get.assert_called_with(api_url, auth=auth)
+        self.assertEqual(result, {})
+
+    @patch('requests.get')
+    def test_get_list_not_found(self, mock_requests_get):
+        listserv_client = ListservClient()
+        response = Bunch(
+            status_code=404,
+            json=lambda: {}
+        )
+        mock_requests_get.return_value = response
+        mailing_list = MailingList.objects.get(id=1)
+        api_url = "%s/%s" % (listserv_client.api_url, mailing_list.address)
+        auth = (listserv_client.api_user, listserv_client.api_key)
+
+        result = listserv_client.get_list(mailing_list)
+
+        mock_requests_get.assert_called_with(api_url, auth=auth)
+        self.assertIsNone(result)
+
+    @patch('requests.get')
+    def test_get_list_exception(self, mock_requests_get):
+        listserv_client = ListservClient()
+        response = Bunch(
+            status_code=500,
+            text="Call to get_list failed."
+        )
+        mock_requests_get.return_value = response
+        mailing_list = MailingList.objects.get(id=1)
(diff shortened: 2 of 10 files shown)


lti_emailer/settings/unit_test.py
```diff
@@ -14,6 +14,12 @@
...
 
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.dummy.DummyCache'
+    },
+}
+
 """
 Unit tests should be able to run without talking to external services like
 redis/elasticache.  Let's disable the caching middleware, and use the db
```
rascalking: BTW, all the stuff below this is only needed if you're using LiveServerTestCase, which you don't seem to be.

manage.py
```diff
@@ -4,7 +4,7 @@
 
 if __name__ == "__main__":
     if 'test' in sys.argv:
-        os.environ.setdefault("DJANGO_SETTINGS_MODULE", "lti_emailer.settings.unit_test")
+        os.environ.update({"DJANGO_SETTINGS_MODULE": "lti_emailer.settings.unit_test"})
```
rascalking: it's just the one value, why not `os.environ['DJANGO_SETTINGS_MODULE'] = 'lti_emailer.settings.unit_test'`?

rascalking:
:+1:

mailing_list/listserv_clients/tests.py
```diff
@@ -0,0 +1,282 @@
...
+            text="Call to get_list failed."
+        )
+        mock_requests_get.return_value = response
+        mailing_list = MailingList.objects.get(id=1)
+        api_url = "%s/%s" % (listserv_client.api_url, mailing_list.address)
+        auth = (listserv_client.api_user, listserv_client.api_key)
+
+        with self.assertRaises(ListservApiError):
+            listserv_client.get_list(mailing_list)
+            mock_requests_get.assert_called_with(api_url, auth=auth)
```
elliottyates: If the `ListservApiError` is raised by the `listserv_client.get_list()` call, it appears this next line will not be reachable. (Same for other `with self.assertRaises()` blocks in these tests.)

mailing_list/models.py
```diff
@@ -111,9 +119,37 @@ def __unicode__(self):
             self.section_id
         )
 
+    def _get_unsubscribed_email_set(self):
+        return {x.email for x in self.unsubscribed_set.all()}
```
elliottyates: I don't see where `unsubscribed_set` is defined -- am I missing something? Does it get generated automatically?

mailing_list/models.py
```diff
@@ -111,9 +119,37 @@ def __unicode__(self):
...
+
+    def _get_enrolled_email_set(self, canvas_course_id):
+        try:
+            canvas_enrollments = get_all_list_data(SDK_CONTEXT, enrollments.list_enrollments_courses, canvas_course_id)
```
elliottyates: Does this also need to be on a section basis, not a course basis?

mailing_list/tests.py
```diff
@@ -0,0 +1,242 @@
...
+                'sapna_mysore@harvard.edu'
+            },
+            add_members={'sapna_mysore@harvard.edu'},
+            delete_members={'david_bonner@harvard.edu'},
+            members_count=3
+        )
+
+    @override_settings(IGNORE_WHITELIST=True)
+    def test_sync_listserv_membership_unsubscribed(self):
+        self._call_sync_listserv_membership(
```
elliottyates: A little unclear what this is doing -- how is this different from `test_sync_listserv_membership_whitelist_not_set()` below?

mailing_list/tests.py
```diff
@@ -0,0 +1,242 @@
...
+                'david_bonner@harvard.edu',
+                'douglas_hall@harvard.edu'
+            },
+            add_members={'david_bonner@harvard.edu'},
+            delete_members=set(),
+            members_count=3
+        )
+
+    def test_sync_listserv_membership_whitelist_not_set(self):
+        self._call_sync_listserv_membership(
```
elliottyates: A little unclear what this is doing -- is this implying that the fourth member (but not the third) in the `enrolled_email_set` is not in the whitelist?

mailing_list/tests.py
```diff
@@ -0,0 +1,242 @@
...
+                'douglashall@g.harvard.edu'
+            },
+            add_members={'david_bonner@harvard.edu'},
+            delete_members=set(),
+            members_count=3
+        )
+
+    @override_settings(IGNORE_WHITELIST=False)
+    def test_sync_listserv_membership_ignore_whitelist_false(self):
+        self._call_sync_listserv_membership(
```
elliottyates: This looks like the same test as `test_sync_listserv_membership_whitelist_not_set` and `test_sync_listserv_membership_unsubscribed`...?

mailing_list/tests.py
```diff
@@ -0,0 +1,242 @@
...
+                'douglashall@g.harvard.edu'
+            },
+            add_members={'david_bonner@harvard.edu'},
+            delete_members=set(),
+            members_count=3
+        )
+
+    @override_settings(IGNORE_WHITELIST=True)
+    def test_sync_listserv_membership_ignore_whitelist_true(self):
+        self._call_sync_listserv_membership(
```
elliottyates: How is this different from `test_sync_listserv_membership_unsubscribed`?

mailing_list/models.py
douglashall: _set is how you access the reverse end of a foreign key relationship with Django models.

mailing_list/listserv_clients/tests.py
douglashall: Good point. I'll remove these.

mailing_list/models.py
douglashall: Good catch.

mailing_list/tests.py
douglashall: There is a model called Unsubscribed that will enable us to allow students to unsubscribe from a mailing list. This is not yet a feature in the UI, but the model will enable such a feature in the future.

'douglas_hall@harvard.edu' is in the Unsubscribed table as defined by the data fixture for the unit tests, so this is testing that Unsubscribed check in sync_listserv_membership.

mailing_list/tests.py
douglashall: Correct. douglashall@g.harvard.edu is not in the whitelist, so this test ensures that this email is not added to the listserv.

mailing_list/tests.py
douglashall: This test makes sure that the IGNORE_WHITELIST setting works as expected.

mailing_list/tests.py
douglashall: The concepts of whitelist and unsubscribed are two different things as commented on above and in the model documentation.

mailing_list/tests.py
elliottyates: Where is the whitelist?

douglashall: The "whitelist" is stored in ml_email_whitelist. There is a data migration that runs when the test database is setup which creates EmailWhitelist models. See mailing_list/migrations/0003_data_add_whitelist_members.py

elliottyates: Ok -- in future I think it would be helpful to call some of this stuff out in the test comments, so people can see pretty quickly why the assertion is expected. E.g. a simple `... # should be in whitelist` in the test data would make it a bit clearer just by reading the test. Otherwise I have to go into multiple files to find whether my assumptions about the test's workings are correct. These tests are pretty straightforward and so don't need many comments in general, as the test names, variables, and structure are quite transparent, but when four different tests have identical nearly identical data and assertions, with small tweaks around overrides and fixture checking, it might help to be clearer as to why a given email is or isn't making it into the assertion list. Thanks though, these tests are clear and easy to build on.

rascalking: Sorry to pile on after I thumbs-upped this, but I have to make my objection to loading data into a table as part of a migration...I didn't notice that's where the whitelist was coming from.  Migrations (imo) are for mapping data from an existing schema into the new schema, not for bulk loading data.  Think something like taking the contents of a text field in a table, populating a new lookup table, and putting the corresponding new id back into the original table.  I'd really expect the whitelist data to live as a fixture for unit testing.

douglashall: Except I want this table populated under normal operations as well, not just unit testing. This table falls under the "populating a lookup table" case.

rascalking: Maybe my explanation wasn't clear, but by populating a new lookup table, I was talking about having code look at the existing values in the database, use those values to populate a new table, then write the corresponding id back to the original table.  Nowhere in the migration (modulo enums) should there be the actual data that goes into the database.

Think of as something that should know how to transform data that exists in an install of the database (and have as few assumptions about that data as possible), not as a way to avoid writing a data loading script/process.

PR #7, opened 2015-04-16
Sprint 20 TLT-1284 Added deployment artifacts

douglashall:
@cmurtaugh 
@bermudezjd

### DIFF
--- a/lti_emailer/settings/base.py
+++ b/lti_emailer/settings/base.py
@@ -142,8 +142,8 @@ USE_TZ = True
 
 STATIC_URL = '/lti_emailer/static/'
 
-REDIS_HOST = SECURE_SETTINGS.get('cache_host', '127.0.0.1')
-REDIS_PORT = SECURE_SETTINGS.get('cache_port', 6379)
+REDIS_HOST = SECURE_SETTINGS.get('redis_host', '127.0.0.1')
+REDIS_PORT = SECURE_SETTINGS.get('redis_port', 6379)
 
 CACHES = {
     'default': {
--- a/lti_emailer/settings/base.py
+++ b/lti_emailer/settings/base.py
@@ -28,6 +28,7 @@ DEBUG = True
 
 ALLOWED_HOSTS = []
 
+ADMINS = SECURE_SETTINGS.get('admins')
 
 # Application definition
 
@@ -141,8 +142,8 @@ USE_TZ = True
 
 STATIC_URL = '/lti_emailer/static/'
 
-REDIS_HOST = SECURE_SETTINGS.get('redis_host', '127.0.0.1')
-REDIS_PORT = SECURE_SETTINGS.get('redis_port', 6379)
+REDIS_HOST = SECURE_SETTINGS.get('cache_host', '127.0.0.1')
+REDIS_PORT = SECURE_SETTINGS.get('cache_port', 6379)
 
 CACHES = {
     'default': {
--- a/lti_emailer/settings/secure.py.j2
+++ b/lti_emailer/settings/secure.py.j2
@@ -1,0 +1,36 @@
+SECURE_SETTINGS = {
+    'canvas_token': '{{ secure_settings.canvas_token }}',
+    'lti_oauth_credentials': {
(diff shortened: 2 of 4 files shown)


lti_emailer/settings/base.py
```diff
@@ -141,8 +142,8 @@
 
 STATIC_URL = '/lti_emailer/static/'
 
-REDIS_HOST = SECURE_SETTINGS.get('redis_host', '127.0.0.1')
-REDIS_PORT = SECURE_SETTINGS.get('redis_port', 6379)
+REDIS_HOST = SECURE_SETTINGS.get('cache_host', '127.0.0.1')
```
cmurtaugh: We currently have secure settings entries for redis_sessions_host and redis_sessions_port; @bermudezjd - what do you think about making these more general like redis_host and redis_port?

lti_emailer/settings/secure.py.j2
```diff
@@ -0,0 +1,36 @@
...
+    'db_default_host': '{{ secure_settings.db_default_host }}',
+    'db_default_port': {{ secure_settings.db_default_port }},
+    'db_termtool_name': '{{ secure_settings.db_termtool_name }}',
+    'db_termtool_user': '{{ secure_settings.db_termtool_user }}',
+    'db_termtool_password': '{{ secure_settings.db_termtool_password }}',
+    'db_termtool_host': '{{ secure_settings.db_termtool_host }}',
+    'db_termtool_port': '{{ secure_settings.db_termtool_port }}',  # Oracle driver requires this to be a string
+
+    'cache_host': '{{ secure_settings.cache_host }}',
+    'cache_port': {{ secure_settings.cache_port }},
```
cmurtaugh: We currently have secure settings defined for redis_sessions_host and redis_sessions_port (as well as redis_cache_host, which includes both the hostname and port). @bermudezjd - what do you think about moving to have just redis_cache_host (with just the hostname) and redis_cache_port?

rascalking: I was going to bring up the same issue with the huey config for this tool.  +1 for a single redis host/port to use from the app.

bermudezjd: That's fine.  I actually created a Trello card to get us away from the redis_sessions plugin and instead set the session backend to `django.contrib.sessions.backends.cache`.  I believe doing so would have the same effect of using redis_sessions and pointing to the same backend (in a different way).  The cache location uses the combined host and port, but we could easily construct that from the split out values.

bermudezjd: Btw, the use of this .j2 file during deployment has been deprecated in favor or reading config from s3.  I'd recommend an example file (secure.py.example) or README doc with the expected secure settings instead.  FWIW, ab_testing has a secure.py.example file.

douglashall: Re: secure.py.j2

Got it. I must have been looking at an out of date play book in tlt-ops. I will remove and add the example file.

douglashall: Re: cache settings

So is it fine to stick with the generic 'cache_host' and 'cache_port' names or do you want them to be redis specific?

bermudezjd: I think generic is fine, although the backend is hardcoded to be redis and I don't see much sense in turning the backend engine into an environmental variable since we should be using the same kind of cache in all of our environments.

bermudezjd: Wait a sec, I missed the fact that the settings vars are called REDIS_* .... I think it would be best to be consistent between the secure keys and the settings where possible.  I'd recommend we update the settings to be CACHE_HOST and CACHE_PORT if we stick with those keys for env variables.

cmurtaugh: Ok - we had a little more discussion in person over here, and I think that a) we should try to keep the key names consistent, and b) I like explicitly mentioning redis in the names in case we decide to add memcached (or some other cache server) to the environment in the future

douglashall: Sounds good. Done.

PR #8, opened 2015-04-16
Sprint 20 TLT-1270 Task queue for mailing list membership sync

rascalking:


lti_emailer/settings/base.py
```diff
@@ -190,6 +191,13 @@
 
 CACHE_KEY_LISTS_BY_CANVAS_COURSE_ID = "mailing_lists_by_canvas_course_id-%s"
 
+HUEY = {
+    'backend': 'huey.backends.redis_backend',
+    'connection': {'host': 'localhost', 'port': 6379},
```
elliottyates: This should probably pull from `SECURE_SETTINGS` in base

mailing_list/tasks.py
```diff
@@ -0,0 +1,57 @@
...
+    filter_kwargs = {}
+    if course_ids:
+        if (not isinstance(course_ids, collections.Iterable)
+                or isinstance(course_ids, basestring)):
+            course_ids = [course_ids]
+        filter_kwargs['canvas_course_id__in'] = course_ids
+
+    success_course_ids, failure_course_ids = [], []
+    for mailing_list in MailingList.objects.filter(**filter_kwargs):
+        success_course_ids.append(mailing_list.canvas_course_id)
```
elliottyates: Don't we want to append this after the try-except block, not before?

lti_emailer/settings/base.py
douglashall: Yes, this can just be:
```
'connection': {'host': REDIS_HOST, 'port': REDIS_PORT},
```

lti_emailer/settings/local.py
```diff
@@ -11,3 +11,7 @@
 DEBUG_TOOLBAR_CONFIG = {
     'INTERCEPT_REDIRECTS': False,
 }
+
+HUEY.update({
+    'connection': {'host': 'localhost', 'port': 6379},
+})
```
douglashall: Looks like this can be removed.

lti_emailer/settings/production.py
```diff
@@ -15,3 +15,11 @@
 
 # Allow the mailing_list app to allow any user to be added to listserv
 IGNORE_WHITELIST = True
+
+# TODO - supply huey.crontab() arguments to determine how often the periodic
+#        mailing list membership sync happens.  defaults to every 1 hour.
+LISTSERV_PERIODIC_SYNC_CRONTAB = {}
+
+# TODO - point huey to the production redis/elasticache server.  we want to
+#        override HUEY['consumer_options'] here.
```
douglashall: Can this be handled via SECURE_SETTINGS in base.py?

lti_emailer/settings/base.py
rascalking: Done.

lti_emailer/settings/local.py
rascalking: Done.

douglashall:
Do we need to run the run_huey management command to start the job workers or does that happen when you start up the Django app? Can we add some basic documentation about how this works to the README?

lti_emailer/settings/production.py
rascalking: Is that where we're putting all the env-specific settings now?  Sure, this was just a placeholder for when we talked to devops about deploying.

douglashall: Yeah, I think that is the idea.

mailing_list/tasks.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@db_periodic_task(
+    crontab(**getattr(settings, 'LISTSERV_PERIODIC_SYNC_CRONTAB',
+                                DEFAULT_LISTSERV_PERIODIC_SYNC_CRONTAB)))
+def periodic_sync_listserv():
+    _sync_listserv(None)
+
+
+def _sync_listserv(course_ids):
```
douglashall: Can we move the contents of this to a utility function that both this and the management command can call?

mailing_list/tasks.py
rascalking: That...is a very good point.  Thanks for the catch.

rascalking:
You have to run the task worker.  I'll add something to the README.

mailing_list/tasks.py
rascalking: Sure, but I *hate* resorting to a `utils.py`.  Any objections to me un-hiding this method and importing it into the management command?

douglashall: Nope. That sounds good.

PR #10, opened 2015-04-21
Feature/update settings for deployment

bermudezjd:
Hardcode default db backend to postgres. Add cookie and environment secure settings. Update canvas api url to be base canvas url for reuse opportunity with other projects. Add a cache key_prefix. Use django cache backend for sessions instead of redis_sessions backend (since cache backend is already redis). Add log_root setting. Add STATIC_ROOT.  Remove gunicorn double requirement.  Simplify other environment setting files. Update secure.py example file.


douglashall:
Looks good! Thanks.

PR #16, opened 2015-04-29
Sprint 21 TLT-1325 Allow students to access a view of the lti_emailer mailing lists

douglashall:
@eparker71 
@elliottyates 

* Note: This PR depends on https://github.com/Harvard-University-iCommons/django-auth-lti/pull/16

### DIFF
--- a/lti_emailer/settings/base.py
+++ b/lti_emailer/settings/base.py
@@ -201,8 +201,10 @@ LISTSERV_API_URL = SECURE_SETTINGS.get('listserv_api_url')
 LISTSERV_API_USER = SECURE_SETTINGS.get('listserv_api_user')
 LISTSERV_API_KEY = SECURE_SETTINGS.get('listserv_api_key')
 LISTSERV_ADDRESS_FORMAT = "canvas-{canvas_course_id}-{section_id}@%s" % LISTSERV_DOMAIN
-LISTSERV_PERIODIC_SYNC_CRONTAB = SECURE_SETTINGS.get('listserv_periodic_sync_crontab',
-                                                     {'minute': '0'})
+LISTSERV_PERIODIC_SYNC_CRONTAB = SECURE_SETTINGS.get('listserv_periodic_sync_crontab', {'minute': '0'})
+
+CACHE_KEY_SECTIONS_BY_CANVAS_COURSE_ID = "sections_by_canvas_course_id-%s"
+CACHE_KEY_ENROLLMENTS_BY_SECTION_ID = "enrollments_by_section_id-%s"
 CACHE_KEY_LISTS_BY_CANVAS_COURSE_ID = "mailing_lists_by_canvas_course_id-%s"
 
 HUEY = {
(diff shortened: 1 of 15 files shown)


lti_emailer/views.py
```diff
@@ -56,4 +59,11 @@ def lti_launch(request):
         "lti_emailer launched with params: %s",
         json.dumps(request.POST.dict(), indent=4)
     )
-    return redirect('mailing_list:index', request.POST['resource_link_id'])
+
+    view = 'mailing_list'
+    if has_lti_role(request, const.LEARNER):
```
rascalking: Doesn't this want to be "if the only lti role is LEARNER"?

mailing_list/api.py
```diff
@@ -41,6 +57,11 @@ def lists(request, resource_link_id):
                 'modified_by': logged_in_user_id
             }
         )
+
+        if has_lti_role(request, const.LEARNER):
```
rascalking: Same as above.

lti_emailer/views.py
douglashall: I think in practice a given user would only have one role, but yes, switching the logic makes sense for when they change their minds and want multiple roles per user. Changing now...

mailing_list/api.py
```diff
@@ -41,6 +57,11 @@ def lists(request, resource_link_id):
...
+            # Learners should only see lists which they can post to, so filter the mailing lists
+            custom_canvas_user_id = int(request.session['LTI_LAUNCH']['custom_canvas_user_id'])
```
rascalking: nitpick: the "custom_" prefix is to namespace is within the lti launch parameters, right?  shouldn't this just be canvas_user_id?

lti_emailer/views.py
elliottyates: Don't you get all roles in Canvas with every LTI launch, or is it just your roles for the course? I think that in the ENV variable Rails provides in the global Javascript context the roles are the set of all roles you have for all courses/accounts, but maybe this is different. If it is the sum total of your roles, this might not be accurate, but if it is only course-level roles, that makes sense.

lti_emailer/settings/base.py
```diff
@@ -201,8 +201,10 @@
 LISTSERV_API_USER = SECURE_SETTINGS.get('listserv_api_user')
 LISTSERV_API_KEY = SECURE_SETTINGS.get('listserv_api_key')
 LISTSERV_ADDRESS_FORMAT = "canvas-{canvas_course_id}-{section_id}@%s" % LISTSERV_DOMAIN
-LISTSERV_PERIODIC_SYNC_CRONTAB = SECURE_SETTINGS.get('listserv_periodic_sync_crontab',
-                                                     {'minute': '0'})
+LISTSERV_PERIODIC_SYNC_CRONTAB = SECURE_SETTINGS.get('listserv_periodic_sync_crontab', {'minute': '0'})
+
+CACHE_KEY_SECTIONS_BY_CANVAS_COURSE_ID = "sections_by_canvas_course_id-%s"
+CACHE_KEY_ENROLLMENTS_BY_SECTION_ID = "enrollments_by_section_id-%s"
```
elliottyates: Maybe an `sdk_` or `canvas_` prefix to make the data source clearer (and distinguish it from the mailer data source)?

rascalking:
:+1:

elliottyates:
:+1:

lti_emailer/views.py
douglashall: The value of the 'roles' parameter in the LTI launch provides the list of roles you have been assigned at the account level, plus the roles you have been assigned for the given course context. I think the logic that Dave has suggested using the roles parameter value is what we want here.

PR #23, opened 2015-05-19
Use CourseSchemaDatabaseRouter to prevent migrations for non-test env…

bermudezjd:
…ironments and don't use any router for unit tests so that migrations are run during testing.  Update versions of canvas sdk and commons_common used in application.  Move icommons_common requirement into base.  Update static url to not include app name prefix.  Remove request log handler and point to log file handler instead.  Allow Django to append `test_` to database name when running unit tests.

@douglashall 
@rascalking 
@elliottyates

### DIFF
--- a/lti_emailer/settings/unit_test.py
+++ b/lti_emailer/settings/unit_test.py
@@ -1,16 +1,13 @@
-from .local import *
+from .base import *
 
 # Make tests faster
+
+DATABASE_ROUTERS = []
+
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': os.path.join(os.path.dirname(__file__), 'test.db'),
-        'TEST_NAME': os.path.join(os.path.dirname(__file__), 'test.db'),
-    },
-    'termtool': {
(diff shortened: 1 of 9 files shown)


elliottyates:
Looks good to me, but don't just take my word for it...

lti_emailer/settings/unit_test.py
```diff
@@ -1,16 +1,13 @@
-from .local import *
+from .base import *
```
rascalking: you're losing ENV_NAME by doing this...do we need to set it to 'unit_test' here?

rascalking:
:+1:

lti_emailer/settings/unit_test.py
bermudezjd: I could do that, although what's the importance of setting ENV_NAME?

rascalking: It's in the other environments, I'd assumed it was there for a reason.  :)

bermudezjd: IIRC, it was used in some emails for the course site wizard to signify which environment was being used.  I think I'll leave it out of here until and unless it becomes useful.

PR #24, opened 2015-06-01
Feature/separate huey logging

bermudezjd:
Add a logfile for huey.  Prefix app log file with "django" to track easier in Splunk.  Cast redis port to int for huey.  Make use of `default_log_level` environment setting to differentiate log settings between deployed environments and projects as needed.

@douglashall


douglashall:
Looks good!

PR #25, opened 2015-06-16
Move KEY_PREFIX for cache into default cache config (where it's suppo…

bermudezjd:
…sed to be).  Also, set use of secure cookies to True since this is the case for all but local environments (and can be true for local envs that use the django ssl server).

### DIFF
--- a/lti_emailer/settings/base.py
+++ b/lti_emailer/settings/base.py
@@ -151,20 +151,16 @@ CACHES = {
         'OPTIONS': {
             'PARSER_CLASS': 'redis.connection.HiredisParser'
         },
-        'TIMEOUT': 60 * 20  # 20 minutes
+        'KEY_PREFIX': 'lti_emailer',  # Provide a unique value for shared cache
+        'TIMEOUT': 60 * 20,  # 20 minutes
     },
 }
-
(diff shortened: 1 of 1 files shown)


douglashall:
Looks good!

elliottyates:
So when this kind of change is made, we'll have to update our secure.py files based on changes to the S3 bucket env variables, yes?

bermudezjd:
@elliottyates: for this particular change I don't think you'd have to update anything.  lti_emailer run locally requires the django ssl server, which means that you should be OK with the requirements of secure cookies on your machine.  I suppose we could just set it to False in local.py if you guys are amenable to that.

elliottyates:
Cool. I know we did away with the example templates so that we could use the S3 stuff as our reference point, so was just thinking about the communication pipeline when S3 stuff changes (new keys, key name changes, removed requirements, etc). It would be good to get a confirmation, when settings files change, of whether or not any changes need to be made to local secure.py. I guess the assumption is "no news is good news", right?

elliottyates:
(I say this because not everyone pays attention to all the updates to all projects all the time, so a stronger heads-up may be required in the case that secure.py needs changes so devs don't miss it. Thanks!)

PR #26, opened 2015-06-22
Sprint 25 TLT-1704 Upgrading internal python package dependencies

douglashall:


lti_emailer/requirements/base.txt
```diff
@@ -11,5 +11,5 @@ django-angular==0.7.13
 dj-static==0.0.6
 huey
 
-git+ssh://git@github.com/penzance/canvas_python_sdk.git@v0.7.6#egg=canvas-python-sdk
-git+ssh://git@github.com/Harvard-University-iCommons/django-icommons-common.git@v1.9.6#egg=django-icommons-common
+git+ssh://git@github.com/penzance/canvas_python_sdk.git@v0.7.7#egg=canvas-python-sdk
```
rascalking: Is it worth moving these down into the other environments so they can follow the same master-vs-tagged model we're using with the others?

rascalking:
:+1:

lti_emailer/requirements/base.txt
douglashall: Yes.

PR #29, opened 2015-06-23
Sprint 25 TLT-1705 Mailgun auth decorator

douglashall:


mailgun/decorators.py
```diff
@@ -0,0 +1,31 @@
...
+
+from functools import wraps
+
+from django.conf import settings
+from django.utils.decorators import available_attrs
+from django.shortcuts import redirect
+from django.core.urlresolvers import reverse_lazy
+
+
+def authenticate(redirect_url=reverse_lazy('mailgun:auth_error'), raise_exception=False):
```
rascalking: If it's not looked at, probably worth removing `raise_exception`.

mailgun/decorators.py
```diff
@@ -0,0 +1,31 @@
...
+    def decorator(view_func):
+        @wraps(view_func, assigned=available_attrs(view_func))
+        def _wrapped_view(request, *args, **kwargs):
+            timestamp = request.POST.get('timestamp')
```
rascalking: Not sure what the spec for the signature is, but typically you want to refuse to verify if the timestamp is outside a certain window.

mailgun/decorators.py
douglashall: Yes.

mailgun/views.py
```diff
@@ -0,0 +1,17 @@
...
+from django.http import HttpResponse
+from django.views.decorators.http import require_http_methods
+
+
+logger = logging.getLogger(__name__)
+
+
+@require_http_methods(['GET'])
+def auth_error(request):
+    return HttpResponse(
```
rascalking: Any reason not to use `[JsonResponse](https://docs.djangoproject.com/en/1.8/ref/request-response/#jsonresponse-objects)`?

mailgun/decorators.py
douglashall: Cool. Should be looking better now.

mailgun/views.py
douglashall: Nope. Changed.

rascalking:
:+1:

PR #30, opened 2015-06-23
Sprint 25 TLT-1672 Mailgun route handler

douglashall:
This adds the Mailgun route handler for notifying sender's of email to mailing lists of which they are not members.


eparker71:
:+1: Look good!

PR #31, opened 2015-06-29
Sprint 25 TLT-1751 Send bounce back response if mailing list is set to readonly

douglashall:
Configuring a mailgun route prevents the default bounce back email we were getting from mailgun when a mailing list is set to readonly from being sent.

This change will have our route handler send a bounce back email to the sender in this scenario.


mailgun/route_handlers.py
```diff
@@ -39,8 +41,22 @@ def handle_mailing_list_email_route(request):
...
-        email_template = get_template('mailgun/non_member_bounce_back.html')
+        logger.info(
+            "Sending mailing list bounce back email to %s for mailing list %s because the sender was not a member",
+            sender,
+            recipient
+        )
+        email_template = get_template('mailgun/email/bounce_back_non_member.html')
+        content = email_template.render(Context({'mailing_list_address': recipient}))
+        subject = "Undeliverable mail sent to %s" % recipient
+        ml.send_mail(sender, subject, html=content)
```
sapnamysore: Minor : Since the 'content = email_template.render..' block is the same and occurs  for both the conditions( in lines 50 and 60), this could  be pulled out side the if -else block and placed outside.

sapnamysore:
Other than a minor suggestion, :+1:

mailgun/route_handlers.py
douglashall: Updated to keep it DRY.

PR #34, opened 2015-07-07
Sprint 26 TLT-1769 Add list of section members, including email address

rascalking:


lti_emailer/canvas_api_client.py
```diff
@@ -21,6 +21,21 @@
 SDK_CONTEXT = SessionInactivityExpirationRC(**settings.CANVAS_SDK_SETTINGS)
 
 
+def get_section(canvas_course_id, section_id):
```
douglashall: Do we need to cache these again individually? Couldn't we just grab them from get_sections below?

mailing_list/models.py
```diff
@@ -122,16 +122,21 @@ def __unicode__(self):
...
             try:
                 univ_ids.append(enrollment['user']['sis_user_id'])
             except KeyError:
-                logger.debug("Found canvas enrollment with missing sis_user_id %s", json.dumps(enrollment, indent=4))
+                logger.debug(
+                    "Found canvas enrollment with missing sis_user_id %s",
+                    json.dumps(enrollment, indent=4))
+        return list(Person.objects.filter(univ_id__in=univ_ids))
 
-        return {p.email_address for p in Person.objects.filter(univ_id__in=univ_ids)}
```
douglashall: Can we keep the original here, returning the generator instead of the fleshed out list?

mailing_list/templates/mailing_list/list_details.html
```diff
@@ -0,0 +1,110 @@
...
+            }
+        }
+
+    }//.fn (end)
+})(jQuery);
+
+$('#people-list').alphabetPagination();
+  </script>
+</body>
+{% endblock body %}
```
douglashall: It would be nice to angularize this as we are moving it over.

mailing_list/templatetags/enrollment_tags.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@register.filter(is_safe=True)
+def enrollment_lname(enrollment):
+    try:
+        lname_first = enrollment['user']['sortable_name']
+    except KeyError:
+        return ''
+    return lname_first.split(',')[0]
+
```
douglashall: These feel like they want to live in django-icommons-common.

mailing_list/templates/mailing_list/list_details.html
```diff
@@ -0,0 +1,110 @@
...
+{% block css %}
+<link rel="stylesheet" type="text/css" href="{% static "harvard-bootstrap/css/harvard-bootstrap.css" %}" media="screen"/>
+{% endblock css %}
+
+{% block body %}
+<body class="lti-tool" role="application">
+  <header>
+    <h1>
+      {% block breadcrumbs %}
+      Course Emailer &gt; {{ section.name | default:section.id }} Members
```
douglashall: "Course Emailer" should be linked to admin_index

lti_emailer/canvas_api_client.py
rascalking: I'm really not a fan of the "grab piles of stuff we might not need, cache it, and hope we use the rest of it later" approach.

mailing_list/models.py
rascalking: Sure.

lti_emailer/canvas_api_client.py
douglashall: We're already using the rest of it though.

mailing_list/templates/mailing_list/list_details.html
rascalking: Thought about trying to move lti_emailer to use ng-route, and angularizing this, but it seemed like more than the story called for.  I figured just get the MVP out there, and we can clean up and/or refactor later.

Ideally, we want to split out the whole UI pattern, put it in icommons-ui, then refactor this and manage sections to use it, but that seemed like a huge pile of scope creep.  Maybe we put in some technical debt to track it?

mailing_list/templatetags/enrollment_tags.py
rascalking: django-icommons-ui, I think...that's where we've got other shared templates, so it's where those templates' custom tags belong too.  Again, putting it there didn't seem to be worth the effort if we weren't also going to refactor manage sections to use it, which felt like scope creep.

mailing_list/templates/mailing_list/list_details.html
rascalking: Good catch.

lti_emailer/canvas_api_client.py
rascalking: In another page, with no guarantee that it'll still be in the cache when this page gets hit.

douglashall: It is highly likely that the sections by course id cache will still be valid when this page gets hit and if not, no worries, that means the other page will load that much faster after people visit this page. I think the point here is that we're caching the same information twice unnecessarily. It would be good to avoid that.

rascalking: Eh, still not a fan of this approach, but I've rewritten it to use that approach, to maintain consistency in the app if nothing else.

mailing_list/templates/mailing_list/list_details.html
rascalking: Split out the JS, called it from app.js.

lti_emailer/settings/base.py
```diff
@@ -194,6 +194,7 @@
 
 MAILGUN_CALLBACK_TIMEOUT = 30 * 1000  # 30 seconds
 
+CACHE_KEY_CANVAS_SECTION_BY_ID = "canvas_section_by_id-%s"
```
douglashall: CACHE_KEY_CANVAS_SECTION_BY_ID can be removed.

douglashall:
Looks good!

PR #35, opened 2015-07-14
Sprint 26 TLT-1797 Added django command to report on users primary email address

douglashall:
```
$ python manage.py report_user_primary_email [term_id]
```

### DIFF
--- a/lti_emailer/management/commands/report_user_primary_email.py
+++ b/lti_emailer/management/commands/report_user_primary_email.py
@@ -80,3 +80,5 @@ class Command(BaseCommand):
             writer.writerow(['Role', 'SIS Email', 'Canvas Email'])
             for role, rows in report_data.iteritems():
                 writer.writerows(rows)
+
+        logger.info("User primary email report complete for term_id %d...", term_id)
--- a/lti_emailer/management/commands/report_user_primary_email.py
+++ b/lti_emailer/management/commands/report_user_primary_email.py
@@ -1,0 +1,85 @@
+import logging
+import csv
+import os
+
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+
+from icommons_common.models import Term, CourseInstance
+
+from canvas_sdk.utils import get_all_list_data
+from canvas_sdk.methods.courses import list_users_in_course_users
+from canvas_sdk.exceptions import CanvasAPIError
+
+from icommons_common.canvas_utils import SessionInactivityExpirationRC
+from icommons_common.models import Person
+
+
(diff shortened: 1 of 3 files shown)


rascalking:
:+1:

lti_emailer/management/commands/report_user_primary_email.py
```diff
@@ -0,0 +1,85 @@
...
+                        list_users_in_course_users,
+                        ci.canvas_course_id,
+                        'email'
+                    )
+                    for user in users_with_email:
+                        user_id = user['sis_user_id']
+                        if user_id not in users:
+                            users[user_id] = {'email': user['email'], 'roles': []}
+
+                    users_with_enrollments = get_all_list_data(
```
elliottyates: Not sure how our SDK supports it, but the following API query gets both enrollments and emails at the same time:
```
courses/6264/users?include[]=email&include[]=enrollments
```

douglashall: Yeah, the answer to that one is that canvas_sdk does not support using multiple include values unfortunately.

lti_emailer/management/commands/report_user_primary_email.py
```diff
@@ -0,0 +1,85 @@
...
+        term_id = options['term_id'][0]
+        try:
+            term = Term.objects.get(term_id=term_id)
+        except Term.DoesNotExist:
+            CommandError('Invalid term_id %d provided.' % term_id)
+        logger.info("Creating Canvas user primary email report for term_id %d...", term_id)
+
+        report_data = {}
+        for ci in CourseInstance.objects.filter(term=term):
+            if ci.canvas_course_id:
```
elliottyates: Could you do something like:
```
ci_list = [ci in CourseInstance.objects.filter(term=term) if ci.canvas_course_id]
for ci in ci_list:
```
This would buy you precious whitespace.

lti_emailer/management/commands/report_user_primary_email.py
elliottyates: Cool. I raised it as an SDK issue. https://github.com/penzance/canvas_python_sdk/issues/50

lti_emailer/management/commands/report_user_primary_email.py
douglashall: But then I would have to iterate the entire list once and then the subset. Whitespace isn't that precious for this one-off report.

elliottyates: :+

PR #36, opened 2015-07-16
Sprint 27 TLT-1826 Remove learner_index, make tool course-admin-only

rascalking:
Also, sort some imports and make some lines fit in 80 columns.


douglashall:
Looks good!

PR #39, opened 2015-07-21
new UI for admin index; also moves more UI elements into icommons-ui

elliottyates:
Attempts to replicate the wireframes for the [admin index](https://projects.invisionapp.com/share/RM3DKOC35#/screens/88963973) and the [access level modal](https://projects.invisionapp.com/share/RM3DKOC35#/screens/88963970). Depends on [django-icommons-ui v1.9.7](https://github.com/Harvard-University-iCommons/django-icommons-ui/pull/64).

@douglashall 
@sapnamysore 
@Vittorio2015 
@roderickmorales


mailing_list/templates/mailing_list/_mailing_list_details.html
```diff
@@ -0,0 +1,147 @@
...
+         ng-class="list.access_level == 'readonly'
+                   ? 'btn-submit-primary-disabled' : 'btn-submit-primary'"
+       {% else %}
+         ng-class="list.access_level == 'readonly'
+                   ? 'btn-submit-secondary-disabled' : 'btn-submit-secondary'"
+       {% endif %}
+       {% verbatim %}
+       ng-disabled="list.access_level == 'readonly'">
+      <i class="fa fa-envelope"></i>
+      {% endverbatim %}
```
douglashall: Can we wrap the whole block in a verbatim tag instead of sprinkling them throughout?

mailing_list/static/mailing_list/js/app.js
```diff
@@ -23,16 +23,29 @@
         ml.isUpdating = false;
         ml.primarySectionLists = [];
         ml.otherSectionLists = [];
+        ml.updatedAccessLevel = '';
```
douglashall: I'm not sure I understand the use of ml.updatedAccessLevel, why not just use the existing property on the list?

elliottyates: If you use the model directly, the underlying page also changes (the values underneath the modal, on the calling page, change as you change the modal). If there's a nicer way to use the model to build the option list and update the backend, but not update the underlying page bindings until the ajax call is successful, I could definitely change this. Perhaps it also merits a comment, since it isn't clear in the code.

mailing_list/templates/mailing_list/admin_index.html
```diff
@@ -2,75 +2,62 @@
...
+    </div>
+
+    <div ng-show="ml.hasOtherSections()">
+      <h3>
+        {% verbatim %}{{ ml.primarySectionLists[0].name }}{% endverbatim %}
+        sections mailing list addresses
+      </h3>
+      <p class="caption-secondary">
+        These sections are created from the
+        <a href="" target="_blank">Manage Sections</a> tool
```
douglashall: Bum link.

elliottyates: Good catch!

mailing_list/static/mailing_list/js/app.js
douglashall: Got it. Makes sense.

douglashall:
There's something funky going on with the font size on the list_details page. Everything is tiny.

roderickmorales:
The tiny text comes from including bootstrap.css

douglashall:
There's a double border going on between section mailing lists on the admin_index page.

mailing_list/templates/mailing_list/_mailing_list_details.html
elliottyates: This because of mixing Django template tag code (`{% if scope == "class" %}`) with Angular code (`{{ list.name }}`). I'm not a fan of all the sprinkling either. I guess I could use Angular includes, but those are slower, no? Would Angular includes fix up this issue (allow you to pass in a similar additional context to the include and conditionally display things)?

elliottyates:
Tiny text (@doug, @roderickmorales) -- good catch, I didn't really test the base.html changes against the list_details page. I'll take a look at that and fix it up.

elliottyates:
Double-border (@doug) -- also a good catch, I didn't even notice it. Will fix up.

mailing_list/templates/mailing_list/_mailing_list_details.html
douglashall: Check out [ng-bind](https://docs.angularjs.org/api/ng/directive/ngBind).

elliottyates: Ah, so do something like `<span ng-bind="list.name"></span>`?

mailing_list/templates/mailing_list/_mailing_list_details.html
```diff
@@ -0,0 +1,147 @@
...
+              <button type="button"
+                      class="close"
+                      data-dismiss="modal"
+                      aria-label="Close">
+                <span aria-hidden="true">&times;</span>
+              </button>
+              <h4 class="modal-title">
+                <label for="permissions-input-{{ list.section_id }}"
+                       class="ic-Label">
+                  Change permissions
```
douglashall: Capitalize "permissions".

mailing_list/templates/mailing_list/admin_index.html
```diff
@@ -2,75 +2,62 @@
...
+  <div class="loading-indicator" ng-show="ml.isLoading">
+    <img src="{% static 'images/ajax-loader-large.gif' %}"/>
+  </div>
+  <main class="ng-hide" ng-show="ml.loaded">
+    <div id="course-sections" ng-show="ml.hasPrimarySections()">
+      <div ng-repeat="list in ml.primarySectionLists">
+        {% verbatim %}
+        <p class="caption-primary" ng-show="$first">
+          <a href="{{ ml.listMembersUrl(list) }}">
+            View course members' email addresses</a>
```
douglashall: Isn't this link redundant with the primary section link below containing the member count?

mailing_list/templates/mailing_list/_mailing_list_details.html
douglashall: Yeah, I would consider that preferable.

mailing_list/templates/mailing_list/admin_index.html
elliottyates: It is -- in fact, the [mockup](https://projects.invisionapp.com/share/RM3DKOC35#/screens/88963973) removes the member count links, and I forgot to remove them. (I double-checked with Vittorio on that one, so if you think the links should stay I'd talk to him.) Thanks.

mailing_list/templates/mailing_list/_mailing_list_details.html
elliottyates: Sure. Uncapitalized text in the [mockup](https://projects.invisionapp.com/share/RM3DKOC35#/screens/88963970) but I think, given the button is capitalized, it makes sense.

mailing_list/templates/mailing_list/admin_index.html
douglashall: Cool. Makes sense.

mailing_list/templates/mailing_list/admin_index.html
```diff
@@ -2,75 +2,62 @@
...
+          ({{ list.members_count }} members)
```
douglashall: Sounds like this should be inside the <a> tag above as well, via the UI/UX team.

PR #40, opened 2015-07-23
prevent mixed content issues with mailto links

elliottyates:
Prevent mixed content warnings and mailto handler redirect failures due to iframe. Logged as bug TLT-1843, apologies for the branch name confusion.

PR #41, opened 2015-07-23
another mailto link fix

elliottyates:
cf #40 .


elliottyates:
Going ahead and merging. Change is identical to #40 .

PR #42, opened 2015-07-23
Task/sapnamysore/tlt 1766 handle custom access level

sapnamysore:
Modifications to add new custom access level. Includes new column in mailing list to track the access level.  Also includes a new router to intercept outgoing emails for lists that are restricted to staff and ensure that the person posting is a member of the CourseStaff table.


lti_emailer/canvas_api_client.py
```diff
@@ -43,6 +43,24 @@ def get_sections(canvas_course_id):
     return result
 
 
+def get_sis_course_id(canvas_course_id):
+    """
+    This method will retrieve the sis_course_id given the canvas course id
+    """
+    print ("in get_sis_course_id for canvas_course_id=%s" %canvas_course_id)
```
douglashall: Remove print.

lti_emailer/canvas_api_client.py
```diff
@@ -43,6 +43,24 @@ def get_sections(canvas_course_id):
...
+        course = response.json()
+        sis_course_id = course['sis_course_id']
+        logger.debug ("returning sis course id  = %s for canvas course %s" % (sis_course_id, canvas_course_id ))
+    except CanvasAPIError:
+        logger.exception(
+            "Failed to get canvas course information for canvas_course_id %s ",
+            canvas_course_id,
+        )
+        raise
+    return sis_course_id
```
douglashall: It would be good to throw the returned canvas course dict into the cache. Check out the other function in this module.

mailing_list/models.py
```diff
@@ -144,6 +154,30 @@ def _get_whitelist_email_set(self):
     def _get_listserv_email_set(self):
         return {m['address'] for m in listserv_client.members(self)}
 
+    def _get_staff_members(self):
+        logger.debug("fetching staff members for canvas course id %s" % self.canvas_course_id)
```
douglashall: This should use the Canvas enrollments API instead of CourseStaff.

mailing_list/models.py
```diff
@@ -151,8 +185,7 @@ def address(self):
             section_id=self.section_id
         )
 
-    @property
-    def access_level(self):
+    def get_listserv_access_level(self):
```
douglashall: We should remove this entirely now that it is a field on the model.

mailing_list/models.py
```diff
@@ -169,13 +202,36 @@ def send_mail(self, to_address, subject='', text='', html=''):
...
+        if it is set to the new custom access of 'staff', set the listserv
+        access level to 'everyone'
+        """
+        #update the db
+        mailing_list = MailingList.objects.get_mailing_list_by_id(self.id)
+        mailing_list.access_level = access_level
+        mailing_list.save()
+
+        listserv_access_level = access_level if access_level != self.ACCESS_LEVEL_STAFF\
+            else self.ACCESS_LEVEL_EVERYONE
```
douglashall: I would go with the following here:
```
if access_level != MailingList.ACCESS_LEVEL_STAFF:
    listserv_access_level = access_level
else:
    listserv_access_level = MailingList.ACCESS_LEVEL_MEMBERS

File: mailing_list/models.py [hunk 6]
```diff
@@ -169,13 +202,36 @@ def send_mail(self, to_address, subject='', text='', html=''):
...
+
+        if access_level == self.ACCESS_LEVEL_STAFF:
+            listserv_access_level = self.ACCESS_LEVEL_EVERYONE
```
douglashall: This if block can be remove.

mailing_list/static/mailing_list/js/app.js
```diff
@@ -37,12 +40,19 @@
             name: {class: 'Disabled', section: 'Disabled'}
         }];
         ml.accessLevelStatus = {
+
```
douglashall: Remove extra whitespace.

mailing_list/models.py
```diff
@@ -34,6 +37,10 @@ def get_mailing_list_by_address(self, address):
             raise MailingList.DoesNotExist
         return MailingList.objects.get(canvas_course_id=canvas_course_id, section_id=section_id)
 
+    def get_mailing_list_by_id(self, id):
+        return MailingList.objects.get(id=id)
```
douglashall: No need for a custom manager method to do this.

mailing_list/models.py
```diff
@@ -169,13 +202,36 @@ def send_mail(self, to_address, subject='', text='', html=''):
...
+        """
+        Since the the listserv_access_level can only be set to one of:
+        'readonly', 'members' or 'everyone', the access level will now be tracked
+        in the DB tsp that we can keep track of the custom access level. The access
+        levels will match for everthing except the custom levels. Eg:
+        if it is set to the new custom access of 'staff', set the listserv
+        access level to 'everyone'
+        """
+        #update the db
+        mailing_list = MailingList.objects.get_mailing_list_by_id(self.id)
```
douglashall: You already have the MailingList instance here with `self`.

mailing_list/models.py
```diff
@@ -169,13 +202,36 @@ def send_mail(self, to_address, subject='', text='', html=''):
...
+        mailing_list.access_level = access_level
+        mailing_list.save()
```
douglashall:
```
self.access_level = access_level
self.save()
```

mailing_list/models.py
```diff
@@ -98,9 +106,11 @@ class MailingList(models.Model):
     ACCESS_LEVEL_MEMBERS = 'members'
     ACCESS_LEVEL_EVERYONE = 'everyone'
     ACCESS_LEVEL_READONLY = 'readonly'
+    ACCESS_LEVEL_STAFF = 'staff'
 
     canvas_course_id = models.IntegerField()
     section_id = models.IntegerField()
+    access_level = models.CharField(max_length=32, blank=True, default='members')
```
douglashall: You should remove the `blank=True` here.

mailing_list/models.py
```diff
@@ -70,7 +77,8 @@ def get_or_create_mailing_lists_for_canvas_course_id(self, canvas_course_id, **k
                 if not listserv_list:
                     listserv_client.create_list(mailing_list)
                 else:
-                    access_level = listserv_list['access_level']
+                    # fetch  access level that is now  stored in the DB
+                    access_level = mailing_list.access_level
```
douglashall: You could remove this else block now and just do `'access_level': mailing_list.access_level` on line 91 below.

mailgun/route_handlers.py
```diff
@@ -43,14 +43,14 @@ def handle_mailing_list_email_route(request):
 
     member_addresses = [m['address'] for m in ml.members]
     bounce_back_email_template = None
-    if ml.access_level == MailingList.ACCESS_LEVEL_MEMBERS and sender not in member_addresses:
+    if ml.get_listserv_access_level == MailingList.ACCESS_LEVEL_MEMBERS and sender not in member_addresses:
```
douglashall: You'll need to make the property change recommended below in order for this to work.

mailgun/route_handlers.py
```diff
@@ -69,3 +69,64 @@ def handle_mailing_list_email_route(request):
...
+    :return:
+    """
+    sender = request.POST.get('sender')
+    recipient = request.POST.get('recipient')
+    subject = request.POST.get('subject')
+    message_body = request.POST.get('body-plain')
+    logger.info("Handling Mailgun outgoing mailing list email from %s to %s", sender, recipient)
+
+    # fetch the mailing list access level
+    mailing_list_access_level = None
```
douglashall: Could you reference ml.access_level below instead of creating this extra variable?

mailgun/route_handlers.py
```diff
@@ -69,3 +69,64 @@ def handle_mailing_list_email_route(request):
...
 
     return JsonResponse({'success': True})
+
+
+
+
+@csrf_exempt
+@authenticate()
+@require_http_methods(['POST'])
+def handle_outgoing_emails_route(request):
```
douglashall: I think that this route handler and the one above will need to be combined and we will need to configure another route 2 routes on mailgun, one that forwards all incoming mailing list email to this new combined router and one that calls a stop() action. We should talk this over more thoroughly in person.

mailing_list/models.py
sapnamysore: I had to remove it as a property when it was  called 'access_level' - as I introduced a new field with the same name. Since it is renamed, can make it a property again

mailing_list/models.py
sapnamysore: Yes, it was on my list to switch it to members instead of everyone.

lti_emailer/canvas_api_client.py
```diff
@@ -59,3 +59,24 @@ def get_enrollments(canvas_course_id, section_id):
...
+    print(" \n\n\n in get_teacher_enrollments for %s" % canvas_course_id)
+    cache_key = settings.CACHE_KEY_CANVAS_TEACHER_ENROLLMENTS_BY_CANVAS_COURSE_ID % canvas_course_id
+    result = cache.get(cache_key)
+    if not result:
+        try:
+            print(" invoking canvas api for  list_enrollments_courses")
+            result = get_all_list_data(SDK_CONTEXT, enrollments.list_enrollments_courses, canvas_course_id,
+                                       type=TEACHER_ENROLLMENT_TYPES)
+            print("\n\n result -------")
+            print(result)
```
douglashall: Let's remove the prints here before merging.

lti_emailer/canvas_api_client.py
```diff
@@ -59,3 +59,24 @@ def get_enrollments(canvas_course_id, section_id):
             raise
         cache.set(cache_key, result)
     return result
+
+
+def get_teacher_enrollments(canvas_course_id):
```
douglashall: Can we use the existing cached enrollments here and filter out the non-"teacher" ones?

sapnamysore: As per an offline discussion, leaving this as is

PR #49, opened 2015-07-31
Feature/aws settings and requirements reorg

bermudezjd:
This PR is centered around the addition of "aws" settings and requirements files to be used during deployment as opposed to environment-specific files.  Some features added are:

* All environment settings files are essentially pointers to a single aws settings file
* All requirements are pinned to a version (including vcs requirements) for consistency among and between servers in an environment, as well as for faster pip installs during deployment
* ALLOWED_HOSTS setting now provides for some level of security now across deployed environments
* Use native redis-py url format for CACHE location
* Introduce secure/env settings for both the default cache timeout and for ignore_whitelist parameter
* Bump some library versions, including pip, django-redis-cache, icommons-common, etc.

@douglashall 
@rascalking


rascalking:
:+1:

PR #50, opened 2015-08-03
Task/sapnamysore/tlt 1846 switch mailto

sapnamysore:
Changes to enable a user to Reply to the individual sending the mail and to include the mailing list if it's a Reply-all.

### DIFF
--- a/mailgun/route_handlers.py
+++ b/mailgun/route_handlers.py
@@ -37,29 +37,20 @@ def handle_mailing_list_email_route(request):
     in_reply_to = request.POST.get('In-Reply-To')
     logger.info("Handling Mailgun mailing list email from %s to %s", sender, recipient)
     if in_reply_to:
-        to_cc_list=[]
+        # If it is a reply to the mailing list, extract the comma/semicolon separated addresses in the To/CC
+        # fields to avoid duplicate being sent
+        logger.debug("This is a reply!! in_reply_to=%s "% in_reply_to)
+        to_cc_list = []
 
         original_to_address = request.POST.get('To')
-        to_cc_list += original_to_address.replace(',', ';').split(';')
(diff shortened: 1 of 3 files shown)


mailgun/route_handlers.py
```diff
@@ -79,13 +84,18 @@ def handle_mailing_list_email_route(request):
         subject = "Undeliverable mail"
         ml.send_mail(sender, subject, html=content)
     else:
-        # Do not send to the sender
+        # Do not send to the sender and also check if it is a reply-all and do not send to original sender to avoid
+        # duplicate being sent as the email client would have already sent it per the Reply-To param set in header
         try:
             member_addresses.remove(sender)
+            if in_reply_to:
+                member_addresses.remove(original_to_address)
```
rascalking: Just a warning that this may be a little fragile...you're assuming they don't rearrange the to/cc fields.  Ideally what we want to do is generate a list of all to/cc addresses, and if any of those are in the list, don't send to them.

mailgun/route_handlers.py
```diff
@@ -79,13 +93,24 @@ def handle_mailing_list_email_route(request):
...
-        # Do not send to the sender
+        # Do not send to the sender. Also check if it is a reply-all and do not send to users in the To/CC
+        # if they are already in the mailing list - to avoid duplicates being sent as the email client would
+        #  have already sent it
         try:
             member_addresses.remove(sender)
+            if in_reply_to:
+                logger.debug("Removing any duplicate addresses =%s from this message as it is a reply all"
+                             % to_cc_list)
+                for item in to_cc_list:
```
rascalking: This can just be ```member_addresses.difference_update(to_cc_list)```.

rascalking:
:+1:

PR #57, opened 2015-08-06
Sprint 28 TLT-1859 TLT-1875 Add "via Canvas" to from field, add "[SHORT TITLE]" to subject field

rascalking:
Also fixes bounce emails.

### DIFF
--- a/mailgun/route_handlers.py
+++ b/mailgun/route_handlers.py
@@ -138,14 +138,21 @@ def handle_mailing_list_email_route(request):
 
         # we want to add 'via Canvas' to the sender's name.  so first make
         # sure we know their name.
+        logger.debug('Original sender name: {}, address: {}'.format(
+                        sender_address.display_name, sender_address.address))
         if not sender_address.display_name:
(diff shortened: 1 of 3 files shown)


mailgun/route_handlers.py
```diff
@@ -91,22 +97,77 @@ def handle_mailing_list_email_route(request):
...
+            logger.debug('Mailgun router handler sending email to {} from {}, '
+                         'subject {}'.format(
+                            member_address, sender_address.full_spec(), subject))
+            ml.send_mail(sender_address.full_spec(), member_address, subject,
+                         text=message_body)
 
     return JsonResponse({'success': True})
+
+
+def _get_name_for_email(canvas_course_id, address):
```
douglashall: This might be better in canvas_api_client.

douglashall:
Looks good.

mailgun/route_handlers.py
rascalking: Good point.

PR #61, opened 2015-08-11
Sprint 28 TLT-1849 Separated course permission check from account permission check

douglashall:
Depends on PR:
https://github.com/Harvard-University-iCommons/django-icommons-common/pull/96


rascalking:
:+1:

PR #63, opened 2015-08-12
Sprint 29 TLT-1880 From address needs to be the mailing list address for maximum deliverability.

douglashall:
Populate Reply-To header with original sender for single replies.


rascalking:
:+1:

PR #67, opened 2015-08-14
HOTFIX v1.2.1 Fixes Canvas permission lookup

douglashall:
The fix is actually in django-icommons-common. This just increments the package version.

https://github.com/Harvard-University-iCommons/django-icommons-common/pull/102


rascalking:
:+1:

PR #68, opened 2015-08-17
Sprint 29 TLT-1892 Mailing list should forward attachments

rascalking:
This is basically 2 changes:
* look for attachments in the router, and use the content-id-map header to figure out which ones are attachments and which are inline
* replace the content-id for inlined images in the plaintext/html body with the content-id we know mailgun will assign on post (they won't let us just specify the original content-it).

### DIFF
--- a/mailgun/route_handlers.py
+++ b/mailgun/route_handlers.py
@@ -195,7 +195,8 @@ def _get_attachments_inlines(request):
         logger.exception('Unable to find content-id map in this email, '
                          'forwarding all files as attachments.')
         content_id_map = {}
-    attachment_name_to_cid = {v: k for k,v in content_id_map.iteritems()}
+    attachment_name_to_cid = {v: k.strip('<>')
+                                  for k,v in content_id_map.iteritems()}
 
     for n in xrange(1, attachment_count+1):
         attachment_name = 'attachment-{}'.format(n)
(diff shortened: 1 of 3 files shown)


mailgun/route_handlers.py
```diff
@@ -161,7 +173,42 @@ def handle_mailing_list_email_route(request):
...
+    attachment_name_to_cid = {v: k.strip('<>')
+                                  for k,v in content_id_map.iteritems()}
+    logger.debug('Attachment name to cid: {}'.format(attachment_name_to_cid))
+
+    for n in xrange(1, attachment_count+1):
+        attachment_name = 'attachment-{}'.format(n)
+        file_ = request.FILES[attachment_name]
+        if attachment_name in attachment_name_to_cid:
+            file_.cid = attachment_name_to_cid[attachment_name]
+            file_.name = file_.name.replace(' ', '_')
```
elliottyates: Any requirement to actually slugify/do more validation, or are spaces the only thing we need to worry about?

elliottyates:
:+1:

elliottyates:
PS if there is mailgun documentation outlining the procedure for specifying "the content-id we know mailgun will assign on post" that might be helpful in line comments (also for the xrange I guess -- your hard-won discoveries are only in the commit comments at the moment).

mailgun/route_handlers.py
rascalking: There's no documentation of what is acceptable to differing email providers...one of outlook/gmail cared about the spaces, the other one didn't.  There's an [RFC](https://tools.ietf.org/html/rfc2392), but the format it describes doesn't match what i've seen generated in the wild by gmail/yahoo/outlook.

rascalking:
Good point...I'll add it in.

PR #69, opened 2015-08-17
Sprint 29 TLT-1896 Use mailgun batch send

rascalking:
This saves us bandwidth, avoids annoying mailgun when we post the same email with 20MB of attachments 300 times, and in spot testing resolves the "attachment is there but empty" problem we'd been seeing with the "one post per recipient" approach.

### DIFF
--- a/mailgun/listserv_client.py
+++ b/mailgun/listserv_client.py
@@ -206,20 +206,23 @@ class MailgunClient(object):
                     len(emails)
                 ))
 
-    def send_mail(self, from_address, reply_to_address, to_address, cc_address,
-                  subject='', text='', html='', attachments=None, inlines=None):
+    def send_mail(self, list_address, from_address, to_address, subject='',
+                  text='', html='', original_to_address=None,
+                  original_cc_address=None, attachments=None, inlines=None):
         api_url = "%s%s/messages" % (settings.LISTSERV_API_URL,
                                      settings.LISTSERV_DOMAIN)
         payload = {
-            'from': from_address,
-            'sender': from_address,
-            'h:Reply-To': reply_to_address,
+            'from': list_address,
+            'h:Reply-To': from_address,
             'to': to_address,
-            'h:cc': cc_address,
             'subject': subject,
             'text': text,
             'html': html
         }
+        if original_to_address:
+            payload['h:To'] = original_to_address
+        if original_cc_address:
+            payload['h:Cc'] = original_cc_address
 
         # we accept a single address or a list of addresses in to_address.
         # if it's a list, add in recipient_variables to make sure mailgun
--- a/mailgun/listserv_client.py
+++ b/mailgun/listserv_client.py
@@ -213,11 +213,12 @@ class MailgunClient(object):
                                      settings.LISTSERV_DOMAIN)
         payload = {
             'from': list_address,
+            'h:List-Id': '<{}>'.format(list_address),
             'h:Reply-To': from_address,
-            'to': to_address,
+            'html': html,
             'subject': subject,
             'text': text,
-            'html': html
+            'to': to_address,
         }
         if original_to_address:
             payload['h:To'] = original_to_address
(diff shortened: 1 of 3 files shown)


mailgun/route_handlers.py
```diff
@@ -156,26 +156,29 @@ def handle_mailing_list_email_route(request):
...
+        logger.debug(
+            "Mailgun router handler sending email to {} from {}, subject {}".format(
+                member_addresses, sender_address.full_spec(), subject
             )
+        )
+        ml.send_mail(
+            sender_address.display_name, sender_address.address,
+            list(member_addresses), subject, text=body_plain, html=body_html,
+            attachments=attachments, inlines=inlines
+        )
```
sapnamysore: Nice! Was it as  simple  as sending a list of addresses? :-)

rascalking: The real key was https://github.com/Harvard-University-iCommons/lti_emailer/pull/69/files#diff-e755209aef70af870f30ef2ff7890ce3R230, which is what tells mailgun not to put that full list of recipients in the final "To" field.

sapnamysore: Yeah, noticed that change too. Was trying to  understand the change there. Glad the doc link is included( as per Elliot's previous suggestion)

mailgun/route_handlers.py
```diff
@@ -156,26 +156,29 @@ def handle_mailing_list_email_route(request):
...
```
sapnamysore: Is it worthwhile having a try-except block around this block to log exceptions in the router?

sapnamysore:
:+1:

mailgun/listserv_client.py
```diff
@@ -221,6 +221,14 @@ def send_mail(self, from_address, reply_to_address, to_address, cc_address,
             'html': html
         }
 
+        # we accept a single address or a list of addresses in to_address.
+        # if it's a list, add in recipient_variables to make sure mailgun
+        # doesn't include the whole list in the to: field, per
+        #   https://documentation.mailgun.com/user_manual.html#batch-sending
+        if not isinstance(to_address, basestring):
```
elliottyates: Do we need to do this for `cc_address` also?

mailgun/route_handlers.py
rascalking: We can, but the default behavior is just going to log it for us anyway.  Might be cleaner to return an explicit error to mailgun, though.

mailgun/listserv_client.py
rascalking: I don't think so...we're sending `cc_address` to mailgun as `'h:cc'`, so mailgun is generally going to ignore the addresses there anyway.

Maybe we want to rename the incoming arg to reflect that usage, but I can't for the life of me come up with a good rename.  You have any ideas?

elliottyates: We talked offline about `h:cc` and a putative `h:to` and whether we can pass those things through on the mails sent out to the individuals on the list -- leaving that in your hands. :)

rascalking: So the end result is we can *almost* replicate the original to/cc fields.  There's no way to prevent mailgun from adding the individual recipient to the To field, but the rest of the to/cc should look the same as it was send.

elliottyates:
Looks good! :+1: 
For someone who isn't fluent in email headers and how mailgun interprets the h:to, h:cc fields, like me, it might take awhile when reading this code for a hypothetical first time to understand why things are the way they are. If we could add comments as to what the h: fields are, and why we send to and cc lists as original_* params to the listserv client, that might help along with the reasons for stripping all that out of the member list, and the purpose of adding the mailing list to the cc field. The code looks great.

elliottyates:
I might add that our growing list of params sent between the interfaces might warrant using a class/object in a future iteration.

rascalking:
Adding comments in before merging down.  And I considered it, but figured we had enough layers of indirection going on here as it was.

elliottyates:
:+1:

PR #70, opened 2015-08-20
Sprint 29 TLT-1902 Add command to find basic mailgun delivery time stats

rascalking:
Output looks like (after debug log messages scroll by)

    Total emails accepted:  85
    Total emails delivered: 85
    Average delivery time:  688 sec
    Max delivery time:      20393 sec

And here's the usage

```
vagrant@precise64:~/lti_emailer(develop) $ python manage.py mailgun_delivery_times --help
usage: manage.py mailgun_delivery_times [-h] [--version] [-v {0,1,2,3}]
                                        [--settings SETTINGS]
                                        [--pythonpath PYTHONPATH]
                                        [--traceback] [--no-color]
                                        [--days-back DAYS_BACK]

Prints average and max delivery time for email sent in the past N days.

optional arguments:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
  -v {0,1,2,3}, --verbosity {0,1,2,3}
                        Verbosity level; 0=minimal output, 1=normal output,
                        2=verbose output, 3=very verbose output
  --settings SETTINGS   The Python path to a settings module, e.g.
                        "myproject.settings.main". If this isn't provided, the
                        DJANGO_SETTINGS_MODULE environment variable will be
                        used.
  --pythonpath PYTHONPATH
                        A directory to add to the Python path, e.g.
                        "/home/djangoprojects/myproject".
  --traceback           Raise on CommandError exceptions
  --no-color            Don't colorize the command output.
  --days-back DAYS_BACK
                        Number of days back to look at emails sent. (default:
                        7)
```

### DIFF
--- a/mailgun/management/commands/mailgun_delivery_times.py
+++ b/mailgun/management/commands/mailgun_delivery_times.py
@@ -1,0 +1,103 @@
+import argparse
+import datetime
+import logging
+import requests
+from collections import defaultdict
+
+import pytz
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+
+
+DATE_FORMAT = '%a, %d %b %Y %H:%M:%S %z'
+EVENTS_MAX_PAGE_SIZE = 300
+EVENTS_OF_INTEREST = ('accepted', 'delivered')
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = ('Prints average and max delivery time for email sent in the past '
+            'N days.')
+
+    def add_arguments(self, parser):
+        parser.add_argument('--days-back', type=int, default=7,
+                            help='Number of days back to look at emails sent. '
+                                 '(default: %(default)s)')
+
+    def handle(self, *args, **options):
+        end = datetime.datetime.now(tz=pytz.utc)
+        begin = end - datetime.timedelta(days=options['days_back'])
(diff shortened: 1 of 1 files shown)


mailgun/management/commands/mailgun_delivery_times.py
```diff
@@ -0,0 +1,103 @@
...
+        except requests.exceptions.RequestException as e:
+            raise CommandError(str(e))
+        events = resp.json()['items']
+
+        # follow 'next' links until we get a blank page
+        while True:
+            url = resp.json()['paging']['next']
+            resp = requests.get(url, auth=auth)
+            try:
+                resp.raise_for_status()
```
sapnamysore: Why is the raise_for_status() done here instead of continuing on with getting the page(line 63) . I am trying to understand why we need this try block

rascalking: `raise_for_status()` throws an exception if we didn't get back a 200.  it's a quick sanity check to make sure the response was happy before moving on with processing the contents.

actually, now that i think of it, my except clause is purposefully broad (catching the base `RequestException` instead of the more specific `HTTPException`), and I'd done that to catch connection/timeout errors.  but that only helps if the `get()` call is inside the try as well...lemme fix that.

sapnamysore:
:+1:

PR #71, opened 2015-08-21
Sprint 29 Get postgres inited and db migrations run on provision

rascalking:
This is basically the same set of changes as https://github.com/Harvard-University-iCommons/icommons_lti_tools/pull/347, but lti_emailer was missing the preactivate/postactivate changes.


eparker71:
:+1:

PR #77, opened 2015-08-27
Sprint 30 TLT-1940 Use UnicodeCSVHandler to output course list details

rascalking:
Since that class assumes unicode inputs, make sure to convert all data in our rows before passing them into writerow().

(includes a fix to vagrant puppet manifest to allow building ndg-httpsclient when running pip install)


douglashall:
Looks good!

PR #79, opened 2015-08-27
Sprint 30 TLT-1888 Delete orphaned mailing lists due to deleted sections

douglashall:


elliottyates:
Looks good, but two questions:
1. Are you bumping icommons_common just to keep the project up-to-date? This doesn't appear to make use of anything past .24 (.24 being a bugfix version).
2. Could you add unit tests for the delete functionality?

douglashall:
@elliottyates 

1. Yes, bumping to keep up to date.
2. I can add a unit test for delete if you think that would add value.

PR #82, opened 2015-08-31
Sprint 30 TLT-1939 Don't put env in tool title if it's "prod"

rascalking:
Tested by setting `ENV_NAME` in `settings/local.py` to a couple different values, including "prod", and verifying we got the tool title we wanted from the tool config page.


douglashall:
Looks good. Some user training around tool configs and the fact that they are editable would be good to.

sapnamysore:
:+1:

PR #83, opened 2015-09-04
Make sure all email address comparisons are lowercased

douglashall:
This addresses a production issue where HKS users were not able to send mail to mailgun mailing list addresses because their email clients were sending their addresses across with capitalized email user names. The fix here is to make sure we lower case all email addresses before comparing them.

### DIFF
--- a/mailgun/route_handlers.py
+++ b/mailgun/route_handlers.py
@@ -49,7 +49,9 @@ def handle_mailing_list_email_route(request):
 
     # if we want to check email addresses against the sender, we need to parse
     # out just the address.
-    sender_address = address.parse(sender)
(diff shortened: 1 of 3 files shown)


rascalking:
@rascalking

mailgun/route_handlers.py
```diff
@@ -49,7 +49,9 @@ def handle_mailing_list_email_route(request):
 
     # if we want to check email addresses against the sender, we need to parse
     # out just the address.
-    sender_address = address.parse(sender)
+    parsed_sender = address.parse(sender)
```
rascalking: Thanks...this is much more legible.

rascalking:
:+1:

PR #85, opened 2015-09-08
Sprint 30 TLT-1961 Remove check for the lis_person_sourceid_param

douglashall:
Sometimes with our test users the `lis_person_sourceid_param` does not get sent in the LTI launch. The fix for this here is to remove the code that was checking for this param because we aren't using the param at all in the view.


eparker71:
:+1: Looks good

PR #86, opened 2015-09-11
made section_id an optional parameter to the list_members view

eparker71:
this was already added to urls.py, it was just missed here in the views.py file.


sapnamysore:
:+1:

PR #89, opened 2015-09-16
minor change to ordering of text under the Sections header of the adm…

elliottyates:
…in index page


elliottyates:
Text change only, tested locally and conforms to Vittorio's expectations. Merging.

PR #94, opened 2015-09-22
Sprint 31 TLT-2006 Cleanly handle emails with attachments and no html body

rascalking:
The logic to correct the content id in the email body assumed we were receiving both html and plaintext formatted bodies.  When that wasn't the case, we tried to `re.sub()` `None`, which throws a `TypeError`.  Let's make sure we're always working with strings for the body variables.


sapnamysore:
Looks good! :+1:

rascalking:
This was merged in from the hotfix/v1.4.2 branch instead.

PR #97, opened 2015-09-28
Sprint 32 TLT-2016 Make sure all log messages use unicode format strings

rascalking:
* also make all usages of logger use the old-style format strings

Tested by deploying to dev, sending email with unicode emdash (ie. `u'\u2014'`) to dev and qa versions of a class mailing list.  The one sent to the dev list came through clean, the one to the qa list didn't.  (verification on splunk pending resolution of current adid login issues)


douglashall:
Looks good!

PR #98, opened 2015-09-29
Sprint 32 TLT-2016 Use unicode format strings for *ALL* logs

rascalking:
* also, convert all logs to '%s'-style format strings


douglashall:
Looks good!

PR #104, opened 2015-09-30
Sprint 32 TLT-1937 Fix unit tests for jenkins and sqlite

rascalking:
* jenkins doesn't have an api key in settings, and unit tests
  shouldn't rely on it being there.  use the same settings override
  used elsewhere in that tests.py.
* generating test schema for sqlite failed due to it adding in an
  extra _id to a field name.  pull in the version of icommons-common
  that includes the fix.


eparker71:
:+1:

PR #107, opened 2015-10-01
Sprint 32 TLT-2037 Optimize send all permission check

douglashall:
* Stop checking a role's permission for every enrollment for every user in a course


rascalking:
@rascalking

rascalking:
:+1:

PR #108, opened 2015-10-01
Sprint 32 TLT-2039 Cache sent message ids

rascalking:
If we take too long to handle a router POST, Mailgun will time out (experiments suggest after 60 secs) and retry later.  If we actually sent the mail along (or sent a bounce email) the first time, then the Mailgun retry will trigger a duplicate email being sent.

This change caches the Message-Id of each email we handle (after sending a bounce, or after sending it to the list members).  It also checks that cache at the top of the router handler, and simply returns success if we know we've handled it before.  This way, when mailgun POSTs a retry, we respond that we've handled it, but we don't send the dupe.

### DIFF
--- a/mailgun/route_handlers.py
+++ b/mailgun/route_handlers.py
@@ -50,9 +50,19 @@ def handle_mailing_list_email_route(request):
 
     attachments, inlines = _get_attachments_inlines(request)
 
-    logger.info(u'Handling Mailgun mailing list email from %s to %s',
-                sender, recipient)
+    logger.info(u'Handling Mailgun mailing list email from %s to %s, '
+                u'subject %s, message id %s',
+                sender, recipient, subject, message_id)
     logger.debug(u'Full mailgun post: %s', request.POST)
+
+    # shortcut if we've already handled this message
+    if message_id:
+        cache_key = settings.CACHE_KEY_MESSAGE_ID_SEEN % message_id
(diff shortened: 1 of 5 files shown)


mailgun/route_handlers.py
```diff
@@ -44,15 +46,26 @@ def handle_mailing_list_email_route(request):
...
     logger.debug(u'Full mailgun post: %s', request.POST)
 
+    # shortcut if we've already handled this message
+    if message_id:
+        cache_key = settings.CACHE_KEY_MESSAGE_ID_SEEN % message_id
+        if cache.get(cache_key):
+            logger.warning(u'Message-Id %s was posted to the route handler, '
+                           u'but we\'ve already handled that.  Dropping.',
+                           message_id)
+        return JsonResponse({'success': True})
```
douglashall: This should be the following, no?
```
if message_id:
    cache_key = settings.CACHE_KEY_MESSAGE_ID_SEEN % message_id
    if cache.get(cache_key):
        logger.warning(u'Message-Id %s was posted to the route handler, '
                                 u'but we\'ve already handled that.  Dropping.',
                                 message_id)
```

rascalking: Yes, yes it should.  Thanks for the catch.

douglashall:
Looks good!

PR #109, opened 2015-10-04
Task/eparker71/tlt 2036 --- WORK IN PROGRESS PR --- there will be more changes

eparker71:
there will most likely be more UI changes and text changes to follow. 
I just wanted to get this out so folks could see the added logic for dealing with SIS sections and Xlisted courses. 

The main areas to look at are:

the get_or_create_or_delete_mailing_lists_for_canvas_course_id method in models.py, utils.py and app.js.

Basically, if a course has an entry in the XList table, it's cross listed and should create the all inclusive course email list. We are no longer relying on the SIS for this. Also, we are able to distinguish between enrollment and non-enrollment sections so we are calling them out in the UI.


mailing_list/templates/mailing_list/admin_index.html
```diff
@@ -64,108 +64,91 @@
...
-                <div class="col-md-12">
-                    <hr width="98%">
-                </div>
-            </div>
+    <div class="col-md-12">
 
-            <!-- otherwise show the primary section -->
-            <div ng-if="ml.primarySectionLists.length == 1" ng-repeat="list in ml.primarySectionLists">
-              {% include "mailing_list/_mailing_list_details.html" with scope="class" %}
+      <!--<hr width="98%">-->
```
sapnamysore: Remove this redundant comment?

mailing_list/templates/mailing_list/admin_index.html
```diff
@@ -64,108 +64,91 @@
...
+                </a>
+              </h4>
             </div>
-            <!-- end show combined mailing list or single primary list -->
 
-            <div ng-show="!ml.hasPrimarySections()" class="col-md-12">
-              <p>
-                  There was an error retrieving the class mailing list for this course.
-                  Please contact your local academic support staff for assistance.
-              </p>
```
sapnamysore: Looks like there was a placeholder for displaying this error message on  not finding primary sections and I see that the hasPrimarySections() method is not there anymore. Do  we not need this check for the method that replaced it?(hasEnrollmentSections?). Maybe not, but am wondering if it was for a specific use case.

PR #110, opened 2015-10-05
Task/eparker71/tlt 2036 course emailer ui changes plus sis section id changes

eparker71:
This PR contains the following

1) text changes

	There are numerous text changes. 
	
2) UI changes

	We now display badges for the different types of sections in the 
	non-enrollment (Email Specific Sections) drop down. 
	
3) logic changes

	This logic is a bit tricky. There are two changes we are dealing with. One is how we
	determine if a course is cross listed and the other is how to deal with the new sections
	being fed by Campus Solutions. 
	
	Main files affected: models.py, utils.py, app.js, admin_index.html

    Unit tests have been updated and added for new methods. 

	Part 1:

	The way we check to see if a course is cross listed has changed. 
	We now check the xreg_map table. If a course has a record in the xreg_map 
	table is it cross listed. There is a new file called utils.py that has a method to
	check for this (is_course_crosslisted). It returns True or False.

	Part 2:

	We now have sections being fed from SIS with section id's that map to records in the 
	course instance table in the course manager database. 

	There is a field in the table called cs_class_type that may contain 1 of 3 
	values ('E', 'N', or None):

	E = This is an enrollment section (not editable by the user)
	N = This is a non-enrollment section (editable by the user in Campus Solutions)
	None = Same as 'E'

	If the sis_section_id exists in the course_instance_table, we check for the values
	above and set the appropriate flag on the section. There is a method in the utils.py
	file called (get_section_sis_enrollment_status) it will return the status based on
	the logic above. 

	We also have sections created by the manage sections tool. These section do not have
	sis_section_id's and will be treated as a non-enrollment sections.


mailing_list/models.py
```diff
@@ -138,15 +137,26 @@ def get_or_create_or_delete_mailing_lists_for_canvas_course_id(self, canvas_cour
                 mailing_list = MailingList(**create_kwargs)
                 mailing_list.save()
 
+            # cs_class_type is used to determine if the section
+            # is an enrollment section or a non-enrollment section.
+            # if it's null for a section with a real sis section id, we
+            # should consider it an enrollment section.
+            cs_class_type = None
+            if s['sis_section_id'] and s['sis_section_id'].isdigit():
```
sapnamysore: Do we only need the check for cs_type for those which have s['sis_section_id'].isdigit()?

mailing_list/static/mailing_list/js/app.js
```diff
@@ -1,134 +1,140 @@
...
-        for (var i = 0; i < ml.accessLevels.length; i++) {
-            var accessLevel = ml.accessLevels[i];
-            ml.accessLevelDisplayNameMap[accessLevel.id] = accessLevel.name;
-            ml.accessLevelDescriptionMap[accessLevel.id] = accessLevel.description;
+    $http.get(URL_LISTS).success(function(data){
+      ml.isLoading = false;
+      var length = data.length;
+      var course_list_member_count = 0;
+      for (var i = 0; i < length; i++) {
+        var list = data[i];
```
sapnamysore: Minor- Would be good to move the list declaration outside the for loop?

mailing_list/static/mailing_list/js/app.js
```diff
@@ -1,134 +1,140 @@
...
+    ml.hasNonEnrollmentSections = function() {
+      return ml.nonEnrollmentSectionLists.length > 0;
+    };
 
-        ml.hasOtherSections = function() {
-            return ml.otherSectionLists.length > 0;
-        };
+    ml.isSisList = function(list){
+      return (list.cs_class_type == 'N');
+    };
```
sapnamysore: Looks like we are checking for type 'N' here. The name is a bit confusing as it would appear we are checking for sis_id - can we rename it ?

sapnamysore: Sorry, I just saw  that these sections are called SIS sections in the UI(but the maintenance of the code may be easier  with more specific names) :-)

mailing_list/templates/mailing_list/admin_index.html
```diff
@@ -62,110 +66,109 @@
...
+            <div class="panel-heading" role="tab" id="headingOne">
+              <h4 class="panel-title">
+                <a class="collapsed"
+                 role="button"
+                 data-toggle="collapse"
+                 data-parent="#accordion"
+                 href="#collapseTwo"
+                 aria-expanded="false" aria-controls="collapseTwo">
+                  <ng-pluralize count="ml.nonEnrollmentSectionLists.length"
+                        when="{'0': 'No Sections', 'one': 'Email Specific Sections', 'other': 'Email Specific Sections'}">
```
sapnamysore: Is the phrase  'Email Specific Sections' intended for both conditions- when length is 1 or greater?  In which case, maybe the check for one is redundant?

sapnamysore: Also , don't we need a numeric check for length? '1' vs one?

mailing_list/models.py
eparker71: there may be a better way to do this, the idea is we will get sis_section_id's that exist in the course instance table those will have a numeric course instance id as well as ones that are strings i.e. "ext:12345". If we got a numeric (or a string that is numeric (is digit = True) we want to see if it exists in the course instance table. If it does, does it have a value for cs_class_type.

mailing_list/static/mailing_list/js/app.js
eparker71: ok, do you mean:
var list;
for(){
list = data[i]
} ?

mailing_list/static/mailing_list/js/app.js
eparker71: yeah, that makes sense will rename.

mailing_list/utils.py
```diff
@@ -0,0 +1,46 @@
...
+    SIS fed sections have an sis_section_id that corresponds to a course_instance_id in the
+    CourseInstance model.
+
+    :param sis_section_id:
+    :return 'N', 'E', or None:
+    """
+    try:
+        ci = CourseInstance.objects.get(course_instance_id=int(sis_section_id))
+        # if there is a course instance but there is no cs_class_type
+        # we should assume this is an enrollment type so return 'E'
```
sapnamysore: With this default condition- every CI section in the system previous to adding the cs_type column will fall into the  'Enrollment' section. Just verifying. Guess only the emailer tool is impacted since we only check it here

sapnamysore:
Looks good! Some minor comments. And a clarification about defaulting the CI sections with none - but looks like those were the requirements.

PR #127, opened 2015-10-13
Sprint 33 TLT-2066 Add support for multiple mailing list addresses in the To field

douglashall:


mailgun/route_handlers.py
```diff
@@ -54,173 +54,175 @@ def handle_mailing_list_email_route(request):
...
-            u'because the sender was not a staff member', sender, recipient)
-        bounce_back_email_template = get_template('mailgun/email/bounce_back_access_denied.html')
-    elif ml.access_level == MailingList.ACCESS_LEVEL_READONLY:
-        logger.info(
-            u'Sending mailing list bounce back email to %s for mailing list %s '
-            u'because the list is readonly', sender, recipient)
-        bounce_back_email_template = get_template('mailgun/email/bounce_back_readonly_list.html')
+    for recipient_address in recipients:
+        recipient = recipient_address.address
+        sender_display_name = parsed_sender.display_name
```
sapnamysore: Can u move line 67 outside the for loop?

douglashall: No, because we modify sender_display_name down below (around 180 or so). We could modify that logic down below to make it a little cleaner, but I'd rather leave it like this now to limit the changes.

sapnamysore: ok cool- seemed like the sender name needn't change for multiple recipients..

sapnamysore:
Looks good :+1:

PR #128, opened 2015-10-14
Sprint 23 TLT-2069 Fix regression in permission check

rascalking:
We were passing in a string where lti_permissions expected a list of strings.  Follow-on change to lti_permissions to catch that case there coming soon.


sapnamysore:
:+1:

PR #132, opened 2015-10-14
added new template for not subscribed messages and updated router to …

eparker71:
In this PR:

added new bounce back template bounce_back_not_subscribed.html to trigger when the 
user is not a member of the list. 

updated text in access denied template based on this doc:
https://docs.google.com/document/d/14uwP0NbY6FPoyz_OB9RA_u5RjFVM8OA1hwsj4_4sYn8/edit

updated mail router to use new template for member not in list error.

### DIFF
--- a/mailgun/templates/mailgun/email/bounce_back_not_subscribed.html
+++ b/mailgun/templates/mailgun/email/bounce_back_not_subscribed.html
@@ -6,9 +6,11 @@
     <body>
         <h4>***Your message could not be delivered to mailing list {{ recipient }}.***</h4>
         <p>
-            The current settings of this mailing list do not allow students and guests to send
-            messages to the list. Teaching staff manage the settings of each mailing list. <a href="https://wiki.harvard.edu/confluence/pages/viewpage.action?pageId=190909873">Troubleshooting tips</a>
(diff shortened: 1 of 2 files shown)


mailgun/templates/mailgun/email/bounce_back_not_subscribed.html
```diff
@@ -0,0 +1,20 @@
+<html>
+    <head>
+        <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+        <meta name="viewport" content="width=device-width">
+    </head>
+    <body>
+        <h4>***Your message could not be delivered to mailing list {{ recipient }}.***</h4>
+        <p>
+            This mailing list does not recognize the email address you are using. This could be because 1) you are not
```
rascalking: I know this matches the text in the google doc, but it seems like it should be `because 1)..., or 2) ...`, or `1) because ..., or 2) because ...`.  The way it is now (`because 1)..., or 2) because ...`) is just awkward.

rascalking:
Otherwise, :+1:

mailgun/templates/mailgun/email/bounce_back_not_subscribed.html
eparker71: good catch!

PR #114, opened 2015-10-07
Feature/selenium lti emailer

hktest:
- Set up of the selenium page-object base structure for the LTI-Emailer Tool
- Added in a simple test to check page loaded.
- Added in a test to masquerade users in different roles and validate access to Emailer tool.
- Added a test_setup.py to add the predefined list of users to any Canvas site (if necessary), for testing against test_permission.py
- Ran each individual tests successfully. Ran regression_suite successfully (when local is set to true or false)

PR #133, opened 2015-10-15
Feature/selenium lti emailer updated

hktest:
Set up of selenium page-object base structure for the LTI-Emailer Tool
Permission tests added to masquerade users in roles to check emailer tool access.  
Ran regression_suite successfully (both locally, and when "run_locally" is set to False)  
***
Note:  
This is a new PR which incorporated feedback from  https://github.com/Harvard-University-iCommons/lti_emailer/pull/114 
New branch since virtually all the files have changed, but keeping that PR open (temporarily) until some folks have a chance to review comments for best practices.


hktest:
@elliottyates @sapnamysore @Joy603

lti_emailer/settings/local.py
```diff
@@ -15,3 +15,14 @@
 }
 
 dictConfig(LOGGING)
+
+CANVAS_URL = 'http://canvas.icommons.harvard.edu/'
```
elliottyates: This isn't necessary -- it comes from `base.py`; to change it, you update your secure settings (see https://github.com/Harvard-University-iCommons/lti_emailer/blob/develop/lti_emailer/settings/base.py#L266). Let's avoid duplication/multiple places to update things.

selenium_tests/lti_emailer/emailer_permission_tests.py
```diff
@@ -0,0 +1,45 @@
...
+    @unpack
+    def test_roles_access(self, user_id, given_access):
+        # This test masquerades as users in roles specified defined in
+        # CANVAS_ADD_USERS (Excel spreadsheet) then validates if the users
+        # are granted/denied access based on their role.
+
+        emailer_main_page = self.emailer_main_page
+        masquerade_page = CanvasMasqueradePageObject(self.driver, self.CANVAS_BASE_URL)
+        masquerade_page.masquerade_as(user_id)
+        masquerade_page.confirm_masquerade()
```
elliottyates: Ah yes, I forgot this part! Good catch. That said, the `confirm_masquerade()` should be _part of_ the `masquerade_as()` function -- in other words, this line should move to the `masquerade_as()` function, at the end, and not be needed in test scripts. (Test scripts shouldn't need to know how masquerade works in Canvas.)

selenium_tests/lti_emailer/page_objects/pin_login_page_object.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        the abstract method can be overridden for individual projects to allow
+        the login to return the appropriate page object for the test. If you do override
+        you will need to call super to invoke the login
+        """
+        self.set_login_type_xid()
+        self.set_username(username)
+        self.set_password(password)
+        self.click_submit()
+        print 'Pin Login page returning Index page for user: %s' % username
```
elliottyates: I'm not sure what this means, or why we need it. What is the Index page? Presumably this can be called from anywhere, and won't necessarily return an 'Index page'. If we're not sure why this is here, let's remove it.

elliottyates:
If the `canvas_masquerade_page_object.py` now lives in the lti_emailer page objects we should remove the canvas/page_objects file hierarchy from the project. (In the future this is the kind of page object that could be used across multiple apps/projects, so we wouldn't always want it to be under an app's test page objects folder.)

selenium_tests/lti_emailer/page_objects/emailer_mainpage_page_object.py
```diff
@@ -0,0 +1,61 @@
...
+
+        if title and 'Course Emailer' in title.get_attribute('textContent'):
+            return True
+        else:
+            raise RuntimeError(
+                'Could not determine if Emailer main page loaded as expected;'
+                'title element was found but did not contain expected text'
+            )
+
+    def get_title(self):
```
elliottyates: See comment re: `test_tool_is_loaded_on_page` -- I think we can remove this method.

elliottyates:
I think we can remove pin/page_objects folders from the project as well. (Same comment there vis-a-vis reusability -- in a project with multiple apps we wouldn't want multiple copies of this. Indeed, we may want to consider moving Canvas page objects and the Pin Login page objects into a library that can be used by all the Selenium tests.)

elliottyates:
There are still some PEP-8 inconsistencies in these files, so before merging let's use the git hook or go over the PyCharm PEP-8 recommendations to fix them up.

lti_emailer/settings/local.py
Joy603: In general,  we should not be checking in these values into GitHub (URLs).  I would expect them to be in the secure.py file along with the user name, password and selenium grid url.

selenium_tests/lti_emailer/page_objects/canvas_masquerade_page_object.py
```diff
@@ -0,0 +1,35 @@
...
+from selenium_tests.lti_emailer.page_objects.emailer_base_page_object import EmailerBasePageObject
+from urlparse import urljoin
+
+
+class CanvasMasqueradePageLocators(object):
+    CONFIRM_BUTTON = (By.CSS_SELECTOR, "a.btn.masquerade_button")
+    MASQUERADE_URL = 'users/{}/masquerade'
+
+
+class CanvasMasqueradePageObject(EmailerBasePageObject):
```
Joy603: Masquerade isn't specific to the lti_emailer, can we put this in base page object?

selenium_tests/lti_emailer/page_objects/emailer_base_page_object.py
```diff
@@ -0,0 +1,23 @@
...
+        try:
+            self._driver.switch_to.frame(self._driver.find_element_by_id("tool_content"))
+        except NoSuchElementException:
+            pass
+
+    def open(self, url):
+        self._driver.get(url)
+        return self
+
+    def find_element(self, *loc):
```
Joy603: find_element and find_elements looks identical to the the ones found in base_page_object, do we really need to re-write them here?

selenium_tests/lti_emailer/emailer_is_loaded_tests.py
```diff
@@ -0,0 +1,22 @@
+import unittest
+
+from selenium_tests.lti_emailer.emailer_base_test_case import EmailerBaseTestCase
+from selenium_tests.lti_emailer.page_objects.emailer_mainpage_page_object import EmailerMainPage
+
+
+class EmailerIsLoadedTests(EmailerBaseTestCase):
+
+    def test_tool_is_loaded_on_page(self):
```
Joy603: I was not expecting this to be a separate file, but an additional tests within emailer_permission_tests.py.  Is the intention to add more is_loaded tests within this file?

selenium_tests/lti_emailer/page_objects/canvas_masquerade_page_object.py
hktest: Good point.  Masquerade is a service. Elliottt originally had it originally structured within a folder -- I see what he was doing there now.  I'll restore that structure (same for PIN service) - thanks!

lti_emailer/settings/local.py
hktest: Thanks, and done. Removed the CANVAS_URL references and added to secure settings instead.  
As for general URL values, I had discussed briefly with Josie. I could also move the emailer_tool_relative_url to secure settings, but the test is dependent on the site setup so the relative url should not be modified.  Before I make another change (I am curious), do URLs really belong to secure_settings? (if they aren't secure settings like username/password?)  What are the general best practices?

elliottyates: @hktest and @Joy603 agreed, it's not black-and-white. I'm ok with reusing as much of the existing base settings as possible -- hence CANVAS_URL coming from secure.py -- but that the tests may have to define their own e.g. relative URLs for now. This goes back to trysting our data is in the same state every time we run a test... I think for now that's our only option. :) Technically the CANVAS_URL probably doesn't _need_ to be in the secure settings, just the base settings, but because it changes in different environments and AWS loads the secure.py settings that's where we pull it from, for convenience. Usernames/passwords/credentials and internal URLs like the selenium grid URL are naturals for including in secure.py, but public URLs probably don't need to be secure (though the argument could be made to put them in a settings file instead of 'baked into' the test files... I just think that's difficult right now given the state of our system and the overhead with making sure everyone's settings point to the right thing). Hope that helps.

rascalking: `secure.py` seems to largely be a misnomer by now.  `secure_and_environment_specific.py` would be a better title.  we've been using it to store settings that vary by environment as well as actual security-sensitive values.

selenium_tests/lti_emailer/emailer_main_page_tests.py
```diff
@@ -0,0 +1,16 @@
...
+from selenium_tests.lti_emailer.page_objects.emailer_mainpage_page_object import EmailerMainPage
+
+
+class EmailerMainPageTests(EmailerBaseTestCase):
+
+    def test_tool_is_loaded_on_page(self):
+        """Check that page is loaded by checking against site title"""
+        driver = self.driver
+        page = EmailerMainPage(driver)  # instantiate
+        page.is_loaded()
```
elliottyates: This isn't asserting anything...? I think it needs to be wrapped in `self.assertTrue(...)`, or else this test will always pass.

selenium_tests/lti_emailer/emailer_permission_tests.py
```diff
@@ -0,0 +1,45 @@
...
+            )
+
+        else:
+            raise ValueError(
+                'given_access column for user {} must be either \'yes\' or '
+                '\'no\''.format(user_id)
+            )
+
+if __name__ == "__main__":
+    unittest.main()
```
elliottyates: New line

selenium_tests/lti_emailer/page_objects/emailer_base_page_object.py
```diff
@@ -0,0 +1,17 @@
...
+    def __init__(self, driver):
+        super(EmailerBasePageObject, self).__init__(driver)
+        try:
+            self._driver.switch_to.frame(self._driver.find_element_by_id("tool_content"))
+        except NoSuchElementException:
+            pass
+
+    def open(self, url):
+        self._driver.get(url)
+        return self
```
elliottyates: New line

selenium_tests/base_page_object.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def focus_on_tool_frame(self):
+        """
+        The pages we are testing are in an iframe, make sure we have the correct focus
+        :return:
+        """
+        try:
+            self._driver.switch_to.frame(self._driver.find_element_by_id("tool_content"))
+        except NoSuchElementException:
+            pass
```
elliottyates: newline

selenium_tests/masquerade/canvas_masquerade_page_object.py
```diff
@@ -0,0 +1,33 @@
...
+from selenium_tests.lti_emailer.page_objects.emailer_base_page_object import EmailerBasePageObject
+from urlparse import urljoin
+
+
+class CanvasMasqueradePageLocators(object):
+    CONFIRM_BUTTON = (By.CSS_SELECTOR, "a.btn.masquerade_button")
+    MASQUERADE_URL = 'users/{}/masquerade'
+
+
+class CanvasMasqueradePageObject(EmailerBasePageObject):
```
elliottyates: Is there some reason this isn't inheriting from the `BasePageObject`, or the `CanvasBasePageObject`? It is outside of the lti_emailer tree structure, so I wouldn't expect it to be dependent upon any of lti_emailer's page objects.

selenium_tests/masquerade/canvas_base_page_object.py
```diff
@@ -0,0 +1,7 @@
+# from selenium_tests.base_page_object import BasePageObject
```
elliottyates: (See comment about `class CanvasMasqueradePageObject` in [canvas_masquerade_page_object.py](https://github.com/Harvard-University-iCommons/lti_emailer/pull/133/files#r42396879)) -- if we're not using this file, it can be removed entirely; but if we're going to use it as a base class for the `CanvasMasqueradePageObject`, this code needs to be uncommented.

selenium_tests/lti_emailer/emailer_main_page_tests.py
hktest: Good point on the assert.  Josie had a similar comment on the existence of this file.  This was just a is_loaded test (basic check to make sure that tool loaded on the page) and not anything else (useful only if the title check does not pass). The key test for this sprint is the checks against  tool_access_permission, so I'm going to remove this file altogether as it is confusing to have it in there to just test is_loaded when there isn't anything I am testing in particular right now for the mainpage.

PR #134, opened 2015-10-15
Sprint 33 TLT-2087 Super Senders

douglashall:
* Adding SuperSender model as a whitelist to allow specified email addresses to send mail to any mailing list in the specified school


rascalking:
:+1:

PR #136, opened 2015-10-19
Sprint 33 Add missing mock to broken-in-jenkins test

rascalking:
Managed to reproduce the breakage on https://jenkins.tlt.harvard.edu/job/lti_emailer-develop/106/console locally by removing `canvas_token` from my `secure.py`.  Adds in a missing mock to the `test_get_or_create_or_delete_mailing_lists_for_canvas_course_id_with_existing_list()` test.  Previously, local tests were succeeding because they were actually calling out to the canvas api.

Also reverts the previously-attempted fix.


elliottyates:
:+1:

PR #137, opened 2015-10-20
short title (of primary course) now appears in the subject line of em…

elliottyates:
…ails to mailing lists in cross-listed courses. This depends on https://github.com/Harvard-University-iCommons/django-icommons-common/pull/128 (`lti_permissions.verification` had the exact logic needed to fix this).

### DIFF
--- a/lti_emailer/requirements/base.txt
+++ b/lti_emailer/requirements/base.txt
@@ -10,6 +10,6 @@ django-angular==0.7.13
 flanker==0.4.34
 ndg-httpsclient==0.4.0
 git+ssh://git@github.com/penzance/canvas_python_sdk.git@v0.8.1#egg=canvas-python-sdk==0.8.1
-git+ssh://git@github.com/Harvard-University-iCommons/django-icommons-common.git@v1.10.8#egg=django-icommons-common==1.10.8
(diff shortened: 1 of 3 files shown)


lti_emailer/requirements/base.txt
```diff
@@ -10,6 +10,6 @@ django-angular==0.7.13
 flanker==0.4.34
 ndg-httpsclient==0.4.0
 git+ssh://git@github.com/penzance/canvas_python_sdk.git@v0.8.1#egg=canvas-python-sdk==0.8.1
-git+ssh://git@github.com/Harvard-University-iCommons/django-icommons-common.git@v1.10.8#egg=django-icommons-common==1.10.8
+git+ssh://git@github.com/Harvard-University-iCommons/django-icommons-common.git@v1.10.8#egg=django-icommons-common==1.10.15
```
eparker71: should be @v1.10.15#egg

eparker71:
:+1:

rascalking:
Might want to add a regression test, now that we're not trying to squeeze it in before the demo.  Otherwise, :+1:

elliottyates:
As discussed offline, the regression test would have to be on the custom manager, which encapsulates all the logic to fix this bug, and doesn't seem worthwhile. Closing and pushing this along.

PR #140, opened 2015-10-22
Task/eparker71/tlt 2087 send email as school liason

eparker71:
## In this PR
* Updated super sender query to allow for mixed case comparison. 
* force all super sender email addresses to be lowercase.
* Adjusted sql for lti_permissions, there was a non visible char at the beginning of each line that prevented postgres from running the query properly.

This ended up being a string case issue. The values in postgres were uppercase for school and a mix of upper and lower for email. This fix will allow for string case mismatch in the db as long as there is a match on the actual data. For example you can now have COLGSAS or colgsas or colGsas in the db and the query will match. Same for email, you can have TLTtest54 or tltTest54 etc...


rascalking:
Pretty sure that was just a tab at the beginning of the sql file...I know I've run that before without issue.

```
$ python manage.py dbshell
Password for user vagrant:
psql (9.4.5)
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-AES256-GCM-SHA384, bits: 256, compression: off)
Type "help" for help.

vagrant=# \i ./mailing_list/data_migrations/lti_permissions.sql
INSERT 0 63
vagrant=#
```

rascalking:
Otherwise, :+1:

PR #143, opened 2015-10-25
Hotfix/v1.5.2

cmurtaugh:
This changes the reply-to address on error emails to be a no-reply address in order to stop bounce-back loops.

### DIFF
--- a/lti_emailer/settings/base.py
+++ b/lti_emailer/settings/base.py
@@ -307,3 +307,5 @@ CACHE_KEY_LISTS_BY_CANVAS_COURSE_ID = "mailing_lists_by_canvas_course_id-%s"
 
 CACHE_KEY_MESSAGE_HANDLED_BY_MESSAGE_ID_AND_RECIPIENT = "lti_emailer:message-handled:%s:%s"
 CACHE_KEY_MESSAGE_HANDLED_TIMEOUT = 60 * 60 * 8  # 8 hours
+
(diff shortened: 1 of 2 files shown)


lti_emailer/settings/base.py
```diff
@@ -307,3 +307,5 @@
 
 CACHE_KEY_MESSAGE_HANDLED_BY_MESSAGE_ID_AND_RECIPIENT = "lti_emailer:message-handled:%s:%s"
 CACHE_KEY_MESSAGE_HANDLED_TIMEOUT = 60 * 60 * 8  # 8 hours
+
+NO_REPLY_ADDRESS = SECURE_SETTINGS.get('no_reply_address', 'no-reply@coursemail.harvard.edu')
```
rascalking: Can't we just construct this based on `LISTSERV_DOMAIN`?  Not a blocker to get a hotfix pushed out, but not my preferred fix.

rascalking:
:+1:

lti_emailer/settings/base.py
cmurtaugh: Well, that'd tie us to using Mailgun addresses as the no-reply addresses (which I used for now because it was expedient); I like the flexibility to be able to have those bounces go somewhere else if necessary.

rascalking: Hrm...surprisingly, they're willing to let us set the from field to a different domain than the sender field...I would have expected them to be too worried about sending reputation to allow that.

https://documentation.mailgun.com/faqs.html#what-is-the-difference-between-the-from-and-sender

Still, if those domains don't match, there's an increased likelihood that email will end up in the spam bucket.

PR #144, opened 2015-10-26
Sprint 34 TLT-2101 Clean up bounce messages

rascalking:
(inadvertently closed out https://github.com/Harvard-University-iCommons/lti_emailer/pull/141 while rebasing on the hotfixes put in over the weekend, sorry)

NOTE: I split out the "handle a route for a single recipient" logic into its own function. Partially this was for sanity sake, but it was also because I noticed us overwriting variables during that logic that would then be used by the next recipient. For example, the subject = title_prefix + ' ' + subject line could run once per recipient, causing the second list in the to/cc fields to have both course titles prepended to its subject.

May be best reviewed in split mode.

### DIFF
--- a/mailgun/route_handlers.py
+++ b/mailgun/route_handlers.py
@@ -95,7 +95,8 @@ def _handle_recipient(request, recipient):
 
     # short circuit if the mailing list doesn't exist
     try:
-        ml = MailingList.objects.get_or_create_or_delete_mailing_list_by_address(recipient)
+        ml = MailingList.objects.get_or_create_or_delete_mailing_list_by_address(
+                 recipient.address)
     except MailingList.DoesNotExist:
         logger.info(
             u'Sending mailing list bounce back email to %s for mailing list %s '
@@ -103,7 +104,7 @@ def _handle_recipient(request, recipient):
         template = get_template('mailgun/email/bounce_back_does_not_exist.html')
         content = template.render(Context({
             'sender': sender,
-            'recipient': recipient,
+            'recipient': recipient.full_spec(),
             'subject': subject,
             'message_body': body_plain or body_html,
         }))
@@ -176,7 +177,7 @@ def _handle_recipient(request, recipient):
         # Send a bounce if necessary
         content = bounce_back_email_template.render(Context({
             'sender': sender,
-            'recipient': recipient,
+            'recipient': recipient.full_spec(),
             'subject': subject,
             'message_body': body_plain or body_html,
         }))
--- a/mailgun/route_handlers.py
+++ b/mailgun/route_handlers.py
@@ -99,8 +99,8 @@ def _handle_recipient(request, recipient):
             u'Sending mailing list bounce back email to %s for mailing list %s '
             u'because the mailing list does not exist', sender, recipient)
         _send_bounce('mailgun/email/bounce_back_does_not_exist.html',
-                     sender, recipient.full_spec(), body_plain or body_html,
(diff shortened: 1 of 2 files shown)


elliottyates:
@elliott

mailgun/route_handlers.py
```diff
@@ -30,193 +30,222 @@
...
+                member_addresses, subject, text=body_plain, html=body_html,
+                original_to_address=original_to_list, original_cc_address=original_cc_list,
+                attachments=attachments, inlines=inlines, message_id=message_id
+            )
+        except RuntimeError:
+            logger.exception(
+                u'Error attempting to send message from %s to %s, originally '
+                u'sent to list %s, with subject %s', parsed_sender.full_spec(),
+                member_addresses, ml.address, subject)
+            raise JsonResponse({'success': False}, status=500)
```
elliottyates: I'm not familiar with `raise`ing an object which is not a subclass of an Exception/Error, so this logic is a little confusing to me, but I guess it's nice to be able to allow for `_handle_recipient()` to throw different types of 'exception responses' if needed in the future and have the `except` clause in the caller be very simple...

mailgun/route_handlers.py
```diff
@@ -30,193 +30,222 @@
...
+            u'Sending mailing list bounce back email to %s for mailing list %s '
+            u'because the mailing list does not exist', sender, recipient)
+        template = get_template('mailgun/email/bounce_back_does_not_exist.html')
+        content = template.render(Context({
+            'sender': sender,
+            'recipient': recipient.full_spec(),
+            'subject': subject,
+            'message_body': body_plain or body_html,
+        }))
+        listserv_client.send_mail(no_reply_address, no_reply_address, sender,
```
elliottyates: Would this happen once for every recipient?

mailgun/route_handlers.py
rascalking: Yeah, there's nothing in the language that requires something you raise to be an exception.  To my mind, it doesn't make sense to define a custom exception class just to embed the response in, and pull it out in the `except` clause.  Especially since this is nominally a private method.

mailgun/route_handlers.py
rascalking: Yes.  If we get an email addressed to two bogus mailing list addresses, we'll send bounces for both addresses.

mailgun/route_handlers.py
```diff
@@ -30,193 +30,222 @@
...
+                             in query.values_list('email', flat=True)}
+
+    # if we want to check email addresses against the sender, we need to parse
+    # out the address from the display name.
+    parsed_sender = address.parse(sender)
+    sender_address = parsed_sender.address.lower()
+    sender_display_name = parsed_sender.display_name
+
+    # any validation that fails will set the bounce template
+    bounce_back_email_template = None
```
elliottyates: Refactoring possibility: Since there is more (and identical?) bounce back code above this, seems like it would be good to have that in a helper function and do a call to it, with a `return` right after, wherever there's a bounce back. Or else raise a custom exception which would encapsulate the listserv_client call perhaps...

elliottyates:
:+1:

mailgun/route_handlers.py
rascalking: Yup.  Previously, we only sent via `listserv_client` in the "not a list" case, so this was as DRY as it could get.  With the hotfixes over the weekend, there's room for more refactoring.

PR #145, opened 2015-10-27
Sprint 34 TLT-2114 Detect and break bounce loops

rascalking:
Adds a check that compares the recipient and from addresses.  If they match, we've got a bounce loop, and we want to drop the email.

Includes a short script used to integration test this.  I made several attempts to provoke the drop via actual email sent through office365 and g.harvard.edu, none of which succeeded.  In the end, I had to resort to doing the POST to the router myself.

### DIFF
--- a/mailgun/route_handlers.py
+++ b/mailgun/route_handlers.py
@@ -35,14 +35,23 @@ def handle_mailing_list_email_route(request):
     '''
     logger.debug(u'Full mailgun post: %s', request.POST)
 
+    from_ = address.parse(request.POST.get('from'))
     message_id = request.POST.get('Message-Id')
     recipients = set(address.parse_list(request.POST.get('recipient')))
-    sender = request.POST.get('sender')
+    sender = address.parse(request.POST.get('sender'))
     subject = request.POST.get('subject')
 
-    logger.info(u'Handling Mailgun mailing list email from %s to %s, '
-                u'subject %s, message id %s',
-                sender, recipients, subject, message_id)
+    logger.info(u'Handling Mailgun mailing list email from %s (sender %s) to '
+                u'%s, subject %s, message id %s',
+                from_, sender, recipients, subject, message_id)
+
+    # short circuit if we detect a bounce loop
+    sender_address = sender.address.lower() if sender else ''
+    from_address = from_.address.lower() if from_ else ''
+    if settings.NO_REPLY_ADDRESS.lower() in (sender_address, from_address):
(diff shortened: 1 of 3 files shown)


elliottyates:
@elliott

mailgun/route_handlers.py
```diff
@@ -82,6 +75,7 @@ def _handle_recipient(request, recipient):
     body_html = request.POST.get('body-html', '')
     body_plain = request.POST.get('body-plain', '')
     cc_list = address.parse_list(request.POST.get('Cc'))
+    from_ = address.parse(request.POST.get('from'))
```
elliottyates: I'm not familiar with this underscore notation, does it signify something?

elliottyates:
(Now that we're also delivering to the sender...) this won't also drop the emails that we send to the sender as a verification that they sent something?

mailgun/route_handlers.py
rascalking: It's the "I want to name a variable X, but X is a reserved word in python, so I have to use X_" syntax.  Which, I'm happily surprised to discover, is actually in pep8, so I'm not just going on what other random people told me.

```
single_trailing_underscore_ : used by convention to avoid conflicts with Python keyword, e.g.

Tkinter.Toplevel(master, class_='ClassName')
```

rascalking:
Nope...emails we're delivering to the list members don't match a route, and thus never get sent to the route handler.

rascalking:
Ugh, though it just occurred to me that we're now sending bounces as `no-reply@coursemail.harvard.edu`, so this comparison won't work any more.

elliottyates:
Sorry got sidetracked, was digging into whether the `mock_ml_get.return_value.send_mail.call_count` mock chains in the unit test were going to resolve properly, they look good.

elliottyates:
Bonus points for the integration test!

elliottyates:
Ok, code looks fine, but it sounds like it may not be the fix we were looking for, based on https://github.com/Harvard-University-iCommons/lti_emailer/pull/145#issuecomment-151849287.

rascalking:
(re: https://github.com/Harvard-University-iCommons/lti_emailer/pull/145#issuecomment-151861188)

Yeah, I've gotten burned by that enough that I was super-paranoid about checking to see it was doing what I thought it was doing.

mailgun/route_handlers.py
elliottyates: Awesome, that's better than `klass` in my opinion.

elliottyates:
:+1:

PR #147, opened 2015-11-04
Adding Jill and Danny to supersender list for HLS

djbrooke:
- To test the supersender changes after the prod release we'll need supersender permission.


eparker71:
:+1:

PR #148, opened 2015-11-05
Sprint 35 TLT-2161 Update the is_primary logic for mailing lists

rascalking:
if there's a course list, we use that, otherwise we use the list from
the primary section.

verified in local testing by running against the root account, manually
inspecting the output to find the cases we might expect:
* courses with no lists (because no sections)
  * eg. course id 5968
* courses with only one list (primary section list is primary)
  * eg. course id 6569
* courses with only one list (no primary section, list in secondary)
  * eg. course id 6721
* courses with a course list and one or more section lists (course list
  is primary)
  * eg. course id 6612
* courses with a primary and one or more non-primary section lists (primary
  section list is primary)
  * eg. course id 6712

### DIFF
--- a/mailing_list/management/commands/activate_all_lists.py
+++ b/mailing_list/management/commands/activate_all_lists.py
@@ -1,5 +1,6 @@
 import argparse
 import logging
+from operator import itemgetter
 
 from django.core.management.base import BaseCommand, CommandError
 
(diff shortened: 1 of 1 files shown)


sapnamysore:
@sapnamysore

mailing_list/management/commands/activate_all_lists.py
```diff
@@ -45,10 +46,18 @@ def handle(self, *args, **options):
                 failures[course['id']] = str(e)
                 continue
 
+            # the "is it the primary list" logic varies depending on whether or
+            # not the list is crosslisted.  figure out which logic to use
+            # before bucketing the lists.
+            if any([l['is_course_list'] for l in lists]):
+                is_primary = itemgetter('is_course_list')
+            else:
+                is_primary = itemgetter('is_primary')
```
sapnamysore: I had to lookup the use of dynamically creating functions using the itemgetter :-). (Have mostly used it in sorting). If the lists has a combination of items containing both 'is_course_list' and ''is_primary' - will  the is_primary function work for both?

sapnamysore:
Other than the clarification, looks good!

mailing_list/management/commands/activate_all_lists.py
rascalking: Yeah, that's why we're picking one based on the presence of any course lists.  If there is one, we want it to be the primary.  If not, we want the primary section to be the primary.

PR #149, opened 2015-11-10
Adds general exception handling to email route handler

elliottyates:
* all failures, including MailingList.send_mail() errors and unhandled errors, now return HTTP 500 response to mailgun with {success:false}, to notify of an error and ask to retry later
* failures also log full POST data from mailgun at level ERROR

### DIFF
--- a/mailgun/route_handlers.py
+++ b/mailgun/route_handlers.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import re
+import sys
 
 from django.conf import settings
 from django.core.cache import cache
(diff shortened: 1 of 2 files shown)


mailgun/route_handlers.py
```diff
@@ -33,45 +34,50 @@ def handle_mailing_list_email_route(request):
...
 
-    logger.info(u'Handling Mailgun mailing list email from %s (sender %s) to '
-                u'%s, subject %s, message id %s',
-                from_, sender, recipients, subject, message_id)
+    except:
+        logger.exception(
+            u'Unhandled exception; aborting. POST data:\n%s\n',
+            json.dumps(request.POST))
+        # tell Mailgun route forwarder to stop processing this message
+        return JsonResponse({'success': False}, status=406)
```
rascalking: Pretty sure we don't want to tell them to stop forwarding it to us.  It's entirely possible the failure could be caused by transient backend issues (ie. db down), and we want them to continue to retry.  I'd say return a 500.

cmurtaugh: Yeah - I think there's little downside to letting MailGun retry.

mailgun/route_handlers.py
```diff
@@ -49,7 +69,7 @@ def handle_mailing_list_email_route(request):
     sender_address = sender.address.lower() if sender else ''
     from_address = from_.address.lower() if from_ else ''
     if settings.NO_REPLY_ADDRESS.lower() in (sender_address, from_address):
-        logger.error(u'Caught a bounce loop, dropping it.  POST data:\n%s\n',
```
rascalking: Leave my double-spaces after a period alone!

mailgun/route_handlers.py
```diff
@@ -24,9 +27,26 @@
 listserv_client = ListservClient()
 
 
+def log_and_report_errors():
```
rascalking: Blegh...this feels like not quite the right name, but I don't have a better one.

mailgun/route_handlers.py
```diff
@@ -64,11 +84,7 @@ def handle_mailing_list_email_route(request):
                                u"for %s, but we've already handled that.  "
                                u'Skipping.', recipient, message_id)
                 continue
-
-        try:
-            _handle_recipient(request, recipient)
-        except JsonResponse as error_response:
```
rascalking: This could be a more specific error, with more helpful error content, that you're throwing away by not catching it.

rascalking:
:+1:

mailgun/route_handlers.py
elliottyates: `mailgun_error_handling`?

rascalking: `handle_uncaught_exceptions`?

mailgun/route_handlers.py
elliottyates: Not following you here -- catching `JsonResponse` just returned the response, so where would the error content be logged if you caught it here and re-`raise`d it? Can you recommend an approach?

mailgun/route_handlers.py
elliottyates: Is this the straw that broke the camel's back that means we'll be implementing a style guide? :)

mailgun/route_handlers.py
rascalking: I was thinking more in terms of the response we're returning might have been more specific.  But I checked, and the one place we raise it, it's just a 500.  So scratch this.

PR #150, opened 2015-11-23
Task/sapnamysore/tlt 2131/logging for webhooks

sapnamysore:
This PR contains changes creating an endpoint that would log  mailgun webhooks post  so that they can easily found in Splunk. More details in JIRA
https://jira.huit.harvard.edu/browse/TLT-2131

Initially I was logging certain keys, but also checked with Devops and they suggested logging the complete data as JSON with some identifiable  prefix. 

Webhooks are currently setup in Dev  for some events and I can see the logs in Splunk with the following query :
tag=dev sourcetype=django_app source="/var/opt/tlt/logs/django-lti_emailer.log" "MAILGUN EVENT"

### DIFF
--- a/mailgun/views.py
+++ b/mailgun/views.py
@@ -2,7 +2,9 @@ import logging
 
 from django.http import JsonResponse
 from django.views.decorators.http import require_http_methods
-
+from django.http import HttpResponse
+from django.views.decorators.csrf import csrf_exempt
+from mailgun.decorators import authenticate
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


mailgun/views.py
```diff
@@ -10,3 +13,18 @@
...
+def log_post_data(request):
+    """
+    This method will help in logging POST data. This method is provided as an endpoint for the
+    Mailgun Webhooks to log an event, so that we can monitor the various configured  events that are available
+    (such as Delivered, Dropped, Bounces, etc)
+    :param request:
+    :return HttpResponse:
+    """
+    logger.info(u'[MAILGUN EVENT] %s', json.dumps(request.POST, indent=4, sort_keys=True))
+    return HttpResponse("Successfully logged post data", status=200)
```
rascalking: Not that it matters, since Mailgun throws it away, but why an HttpResponse instead of the already-imported JsonResponse?

rascalking:
:+1:

mailgun/views.py
sapnamysore: No reason really. Just kept it simple and used HttpResponse , didn't notice the JsonResponse

mailgun/views.py
```diff
@@ -10,3 +13,18 @@
 @require_http_methods(['GET'])
 def auth_error(request):
     return JsonResponse({'error': 'Failed to authenticate request.'}, status=401)
+
+
+@csrf_exempt
+@authenticate()
+@require_http_methods(['POST'])
+def log_post_data(request):
```
elliottyates: If this was called `log_webhook()` it might be clearer what it's for (i.e. which post data it's collecting).

sapnamysore: I intentionally left it generic so it could be used to log  any post info as there isn't anything specific to webhooks here..

elliottyates:
:+1:

PR #151, opened 2015-11-30
Sprint 36 TLT-2160 Use correct display name for sender

rascalking:
Mailgun seems to only use the email address for the `sender` field in their POST, so looking for its display name isn't ever going to work.  Instead, assuming the `sender` and `from` fields agree on the email address, let's use the `from` display name.  If that's missing, then fall back on enrollee lookup.

### DIFF
--- a/mailgun/tests.py
+++ b/mailgun/tests.py
@@ -25,9 +25,11 @@ class RouteHandlerUnitTests(TestCase):
 
     def setUp(self):
         self.factory = RequestFactory()
-        self.user = User.objects.create_user(username='unittest',
-                                             email='unittest@example.edu',
-                                             password='unittest')
+        self.user = User.objects.create_user(email='unittest@example.edu',
+                                             password='insecure',
+                                             username='unittest')
+        self.user.first_name = 'Unit'
+        self.user.last_name = 'Test'
 
     @override_settings(CACHE_KEY_MESSAGE_HANDLED_BY_MESSAGE_ID_AND_RECIPIENT='%s:%s')
     @patch('mailgun.route_handlers.cache.get')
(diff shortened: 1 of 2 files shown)


elliottyates:
@elliottyates

mailgun/tests.py
```diff
@@ -261,6 +265,133 @@ def test_bounce_loop_detection(self, mock_ml_get, mock_ci_get, mock_ss_filter, m
...
+    @patch('mailgun.route_handlers.get_name_for_email')
+    @patch('mailgun.route_handlers.SuperSender.objects.filter')
+    @patch('mailgun.route_handlers.CourseInstance.objects.get_primary_course_by_canvas_course_id')
+    @patch('mailgun.route_handlers.MailingList.objects.get_or_create_or_delete_mailing_list_by_address')
+    def test_sender_display_name_from_field(self, mock_ml_get, mock_ci_get, mock_ss_filter,
+                                            mock_get_name_for_email):
+        '''
+        TLT-2160
+        Verifies that if the mailgun `sender` field lacks a display name, we'll
+        use the one in the mailfun `from` field before resorting to enrollee
```
elliottyates: 'mailfun' -- Freudian slip?

elliottyates:
Looks good. :+1: Does the parsing library handle left/right-whitespace padding? If not, do we want to be paranoid and `strip()` the display name strings before checking for `None`? Or just let their display name indeed be possibly blank if they've configured their client that way?

mailgun/tests.py
rascalking: I think a Freudian slip would not have substituted `fun`.  :)

rascalking:
Yeah, it's my understanding that they handle excess leading/trailing whitespace in the display name.

Hah, forgot they provide a grammar for their parser:
https://github.com/mailgun/flanker/blob/master/docs/User%20Manual.md#grammar

PR #153, opened 2015-12-01
Task/josie/selenium task

Joy603:


selenium_tests/lti_emailer/page_objects/emailer_mainpage_page_object.py
```diff
@@ -55,3 +57,16 @@ def is_loaded(self):
...
+        return element
+
+    def select_member_link(self):
+        """
+        select the course info link element and click it
+        :returns CourseSearchPageObject
+        """
+        self.focus_on_tool_frame()
+        self.find_element(*EmailerMainPageLocators.MEMBER_LINK).click()
+        return CourseEmailerListPage(self._driver)
```
rascalking: This may be just a personal preference thing, but having page objects need to know about other page objects seems like it's going to get messy (it can lead to circular imports, at least).  Why not rely on the test to know that it's transitioned to another page, and to instantiate the new page object there?

selenium_tests/lti_emailer/page_objects/emailer_member_list_page_object.py
```diff
@@ -0,0 +1,32 @@
...
+    def is_loaded(self):
+        """
+        Verifies that the page is loaded correctly by validating the title
+        :returns True if title matches else a RuntimeError is raised
+        """
+        # Note: this just checks that the  title is displayed;
+        # it doesn't guaranteed that everything we expect is rendered on the
+        # page, because angular fetches the data asynchronously
+
+        title = None
```
rascalking: Nitpick: don't need this line.  If there's an exception getting the title, we return immediately.  Otherwise, it's defined inside the `try`.

selenium_tests/lti_emailer/regression_suite.py
```diff
@@ -2,31 +2,36 @@
...
-# create a test suite combining the tests above
-smoke_tests = unittest.TestSuite([permission_testing])
+        permission_testing = unittest.TestLoader().loadTestsFromTestCase(EmailerPermissionTests)
 
-# run the suite
-runner.run(smoke_tests)
-# close test report file
-report_file_buffer.close()
+        # create a test suite combining the tests above
+        smoke_tests = unittest.TestSuite([permission_testing])
```
rascalking: Same comment about auto-discovering the tests as in the other PRs applies here.

rascalking:
:+1:

selenium_tests/lti_emailer/page_objects/emailer_member_list_page_object.py
Joy603: Removed.

selenium_tests/lti_emailer/regression_suite.py
```diff
@@ -2,31 +2,36 @@
 import time
 from os import path, makedirs
 
-import HTMLTestRunner
+from selenium_common import HTMLTestRunner
 from selenium_tests.lti_emailer.emailer_permission_tests import EmailerPermissionTests
 
 
-date_timestamp = time.strftime('%Y%m%d_%H_%M_%S')
+class RegressionSuite(unittest.TestCase):
```
rascalking: Forgot to ask...why is this now a `TestCase`?  And why is all the logic now in its `main` function?

selenium_tests/lti_emailer/page_objects/emailer_mainpage_page_object.py
Joy603: Fixed

selenium_tests/lti_emailer/regression_suite.py
Joy603: The auto-discovering will make sense when you don't care about the order that it finds and execute the tests.

rascalking: Do we care about the order that it runs the tests in now?

Joy603: Not yet, but just something to keep in mind.

selenium_tests/lti_emailer/regression_suite.py
Joy603: I think I did this for performance improvement purposes. Maybe I should rename the 'main' function?

rascalking: I don't know where `main` came from, but it's not a standard `unittest.TestCase` method.  Mostly I don't understand why we're wrapping logic to bring up a loader, generate a test suite, and run tests, all in a class designed to have individual test cases on it.

Joy603: Fixed

PR #154, opened 2015-12-03
Sprint 37 TLT-2108 Handle missing attachments cleanly

rascalking:
Notice missing attachments in the POST, log the fact that it was missing, and return a 400 error to mailgun.

NOTE: This adds the capability to raise an encapsulated HttpResponse from within the route handler, and to have it returned.  Also includes a unit test for that functionality.

### DIFF
--- a/mailgun/exceptions.py
+++ b/mailgun/exceptions.py
@@ -1,3 +1,20 @@
+from django.http import HttpResponse
+
+
 class HttpResponseException(RuntimeError):
+    '''Encapsulates a django.http.HttpResponse.  Intended to be caught at the
+    top level of a view (or by a view decorator) to allow code within a view
+    to immediately return a response.'''
+
     def __init__(self, response):
+        if not isinstance(response, HttpResponse):
+            raise TypeError(
+                      u'HttpResponseException expected an HttpResponse, got'
+                      u'a {}'.format(type(response)))
         self.response = response
+
+
+    def __unicode__(self):
+        return u'HttpResponseException<{} {}>: {}'.format(
+                   type(self.response), self.response.status_code,
+                   unicode(self.response))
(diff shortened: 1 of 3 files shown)


mailgun/exceptions.py
```diff
@@ -0,0 +1,3 @@
+class HttpResponseException(RuntimeError):
```
elliottyates: Why is this called `HttpResponseException`? It seems to always encapsulate JSON responses (to mailgun), and I can't see any references to the new `HttpResponse` imports in route_handlers.py or tests.py. Please educamate me.

mailgun/exceptions.py
```diff
@@ -0,0 +1,3 @@
...
+    def __init__(self, response):
+        self.response = response
```
elliottyates: Would we like to have `__str__()` and `__unicode__()` methods for this custom object?

elliottyates:
:+1:

mailgun/exceptions.py
rascalking: The intent is for it to encapsulate any `django.http.HttpResponse` subclass (which I should probably be asserting on init, actually), hence the name.  It's there largely because I miss flask's `abort()` method (okay, and because my understanding of `raise` was wrong, it won't let me raise an instance of a new-style object that doesn't derive from `Exception`).  There doesn't seem to be a builtin way in django for you to just short-circuit and return an HttpResponse from deep in the call stack in your view.  Which, given that that's the point where you know specifically the type and details of the response you want, seems silly.  

An alternative design is to create a collection of HttpException objects, subclassed per status code, and catch *that* in the handler, construct the response, and return it.  That is probably slightly cleaner (in that the thing throwing the error doesn't have to know the format of the response you're expected to return), but it felt like overkill.  Let me know if you strongly prefer that approach.

mailgun/exceptions.py
rascalking: Sure.

PR #155, opened 2015-12-07
Sprint 37 TLT-2108 Add integration test script

rascalking:
cut-pasted from the one for 2114, with the post body and response
handling tweaked.  next time we want one of these, i library it up good.

* also rename integration test for 2114, fix typo in it

### DIFF
--- a/mailgun/test_scripts/tlt-2108-handle-missing-attachments-integration-test.py
+++ b/mailgun/test_scripts/tlt-2108-handle-missing-attachments-integration-test.py
@@ -52,5 +52,10 @@ if env_name not in ('dev', 'qa', 'stage'):
 # post it
 url = 'https://lti-emailer.%s.tlt.harvard.edu/mailgun/handle_mailing_list_email_route/' % env_name
 resp = requests.post(url, data=post_body)
-if not resp.status_code == 400:
+
+# we're expecting a 400 client error, since we sent a malformed POST.  if we
+# didn't get that, squawk about it.
+if resp.status_code != 400:
     resp.raise_for_status()
+
+# at this point, check splunk for 'Attachment "missing from POST"'
--- a/mailgun/test_scripts/tlt-2108-handle-missing-attachments-integration-test.py
+++ b/mailgun/test_scripts/tlt-2108-handle-missing-attachments-integration-test.py
@@ -34,11 +34,12 @@ def generate_signature_dict():
 # prep the post body
 list_address = 'canvas-6760-2069@mg.dev.tlt.harvard.edu'
 post_body = {
-    'sender': 'Integration Test <integrationtest@example.edu>',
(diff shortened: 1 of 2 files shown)


mailgun/test_scripts/tlt-2108-handle-missing-attachments-integration-test.py
```diff
@@ -0,0 +1,56 @@
...
+
+# figure out which server to post to
+env_name = settings.ENV_NAME
+if env_name not in ('dev', 'qa', 'stage'):
+    env_name = 'dev'
+
+# post it
+url = 'https://lti-emailer.%s.tlt.harvard.edu/mailgun/handle_mailing_list_email_route/' % env_name
+resp = requests.post(url, data=post_body)
+if not resp.status_code == 400:
```
elliottyates: Why not `400`?

elliottyates: Is that because it's the one expected?

elliottyates: In which case, could these scripts output something, and/or have brief comments or `print()` statements/stdout logging that would explain what they're testing and what the output should be?

rascalking: Yup.  As to why I didn't write this as `if resp.status_code != 400:`, that's beyond me.

elliottyates:
:+1:

mailgun/test_scripts/tlt-2108-handle-missing-attachments-integration-test.py
rascalking: And sure, though I was going for the usual unix-y "be quiet unless something fails".

elliottyates: That makes sense in the context of an automated script.

rascalking: Not just, though.

elliottyates: :) That's a matter of preference, I guess. If QA is running these things, my thought is they'd prefer verbose output. As long as we're all on the same page. If they're going to act like a unix command, maybe the comments about how they're used and what to expect could just be code comments. I don't care that much, but what it means to act unix-y isn't something everyone knows about, or even those who do (like me) don't necessarily assume scripts work like that. I'm usually much better at asking others for documentation than writing it myself, it must be added.

rascalking: Yah, I opted for adding comments in the code, including what to look for in splunk.

PR #159, opened 2016-03-25
Sprint 44 TLT-1310 Exit regression suite with error on test failure, some fixes

rascalking:
This change also bumps the versions of selenium and selenium_common,
with corresponding fixes to settings and some selenium tests to attempt
to get the tests to pass.  At this point they're still failing, but due
to a nonsense-looking tool_id being used.  Since this change was just
supposed to get failures to return an error code, and they're now doing
that, i'm calling it a day here.


eparker71:
:+1:

PR #161, opened 2016-04-05
Sprint 45 TLT-2306 Send dummy body to mailgun if there is no body

rascalking:
Mailgun's API may not document it, but it won't send emails for you that have empty values for both the `body-plain` and `body-html` parameters.  But they will send emails with `'body-plain': ' '` (ie. a single space for a plain body).  So detect the empty body case, and work around it.

NOTE: this change includes a test script, since neither outlook nor gmail's web client would send emails with no body at all (they both supplied an empty html document for `body-html`).

### DIFF
--- a/mailgun/test_scripts/tlt-2306-empty-bodies-integration-test.py
+++ b/mailgun/test_scripts/tlt-2306-empty-bodies-integration-test.py
@@ -41,7 +41,6 @@ base_post = {
     'sender': 'Integration Test <integrationtest@example.edu>',
     'from': settings.NO_REPLY_ADDRESS,
     'recipient': list_address,
-    'subject': 'blah',
     'To': list_address,
 }
 base_post.update(generate_signature_dict())
@@ -57,6 +56,7 @@ posts = []
 for body in bodies:
     post = copy.deepcopy(base_post)
     post.update(body)
+    post['subject'] = 'empty body test where body="{}"'.format(json.dumps(body))
     posts.append(post)
 
 # figure out which server to post to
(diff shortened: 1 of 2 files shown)


mailgun/test_scripts/tlt-2306-empty-bodies-integration-test.py
```diff
@@ -0,0 +1,76 @@
...
+
+import requests
+from django.conf import settings
+
+try:
+    settings.NO_REPLY_ADDRESS
+except ImportError:
+    # make sure the project root is in sys.path
+    import os, sys
+    sys.path.insert(0, os.path.normpath(
```
elliottyates: wow

mailgun/test_scripts/tlt-2306-empty-bodies-integration-test.py
```diff
@@ -0,0 +1,76 @@
...
+    {},
+    {'body-html': '', 'body-plain': ''},
+    {'body-html': ''},
+    {'body-plain': ''},
+]
+posts = []
+for body in bodies:
+    post = copy.deepcopy(base_post)
+    post.update(body)
+    posts.append(post)
```
elliottyates: Would it be helpful to make the subject of the email a string repr of the `body`, e.g. 'testing case where body={...}' ?

elliottyates:
:+1:

elliottyates:
How are we tracking these tests? They are run-once by dev and QA in dev,qa,stage, then we release into the wild and never touch them again, yes?

mailgun/test_scripts/tlt-2306-empty-bodies-integration-test.py
rascalking: Yup.

rascalking:
Yeah...they're there because there's no easy way to craft these emails short of doing these scripts...they're not intended to be automation.

PR #162, opened 2016-04-07
Sprint 45 TLT-2306 Fix integration test

rascalking:
* Get the environment name earlier, and use it in the list address.
* Don't use no-reply as the from address, it will trigger our bounce loop dropper.


elliottyates:
:+1:

PR #167, opened 2016-07-18
TLT-1943 - mailing list option to not include all staff on section lists

elliottyates:


mailing_list/migrations/0013_auto_20160715_1841.py
```diff
@@ -0,0 +1,34 @@
...
+        ('mailing_list', '0012_supersender'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='CourseSettings',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('canvas_course_id', models.IntegerField()),
+                ('alwaysMailStaff', models.NullBooleanField(default=True)),
```
eparker71: should probably make this 'always_mail_staff' instead of 'alwaysMailStaff'

eparker71:
I made a change to the name of the staff flag in the MailingList model. I changed alwaysMailStaff to always_mail_staff to be more consistent with the naming scheme of the project. I updated files and tests to reflect this change.

eparker71:
@elliottyates 👍 on the UI changes!!!

mailgun/route_handlers.py
```diff
@@ -141,11 +141,22 @@ def _handle_recipient(request, recipient):
...
+
+    # conditionally include staff addresses in the members list. If always_mail_staff is true
+    # all staff will receive the email
     teaching_staff_addresses = ml.teaching_staff_addresses
-    member_addresses = teaching_staff_addresses.union(
+    if ml.course_settings and ml.course_settings.always_mail_staff:
+        member_addresses = member_addresses.union(teaching_staff_addresses)
+    elif ml.course_settings == None:
+        member_addresses = member_addresses.union(teaching_staff_addresses)
+        # if the course settings object does not exist create it, this equates to True
```
elliottyates: It looks like this code makes assumptions about the default value for the model field `always_mail_staff`. It's probably better to just rely on that value entirely, so that any change to the model will automatically be handled here without having to remember that there's a dependency.

So perhaps something like:

```
# if the course settings object does not exist, create it to initialize the defaults
if ml.course_settings == None:
    ml.course_settings = CourseSettings.objects.create(canvas_course_id=ml.canvas_course_id)
    ml.save()

if ml.course_settings and ml.course_settings.always_mail_staff:
    member_addresses = member_addresses.union(teaching_staff_addresses)
```

This is also a little DRYer.

mailgun/route_handlers.py
```diff
@@ -141,11 +141,22 @@ def _handle_recipient(request, recipient):
...
+        ml.course_settings = CourseSettings.objects.create(canvas_course_id=ml.canvas_course_id)
+        ml.save()
+
+    # create a list of staff plus members to use to check permissions against
+    # so all staff can email all lists
+    staff_plus_members = teaching_staff_addresses.union(
```
elliottyates: We probably want to reuse `member_addresses`, defined already on line 144, to avoid extra computation on line 160.

mailgun/tests.py
```diff
@@ -269,20 +269,156 @@ def test_empty_body_html(self, mock_ml_get, mock_ci_get, mock_ss_filter):
     @patch('mailgun.route_handlers.SuperSender.objects.filter')
     @patch('mailgun.route_handlers.CourseInstance.objects.get_primary_course_by_canvas_course_id')
     @patch('mailgun.route_handlers.MailingList.objects.get_or_create_or_delete_mailing_list_by_address')
-    def test_multi_mailing_list_recipients(self, mock_ml_get, mock_ci_get, mock_ss_filter):
+    def test_multi_mailing_list_recipients_with_course_settings_none(self, mock_ml_get, mock_ci_get, mock_ss_filter):
```
elliottyates: Maybe we can update the docstring to describe the importance of the `_with_course_settings_none`?

We may not need all this anyway. `test_multi_mailing_list_recipients_with_always_mail_staff_true()` checks the behavior when `always_mail_staff==True`, so perhaps this could either reuse that test or simply check that `always_mail_staff` is indeed set to `True` when you send mail to a list that doesn't have a `CourseSettings` object...? Just want to avoid lots of duplication, and minimize the number of tests that check what the field defaults mean in the real world.

mailgun/tests.py
```diff
@@ -269,20 +269,156 @@ def test_empty_body_html(self, mock_ml_get, mock_ci_get, mock_ss_filter):
...
+            original_cc_address=[],
+            original_to_address=[u'class-list@example.edu', u'bogus@example.edu'],
+            text=u'blah blah'
+        )
+        ml.send_mail.assert_has_calls([send_mail_call, send_mail_call])
+
+    @patch('mailgun.route_handlers.SuperSender.objects.filter')
+    @patch('mailgun.route_handlers.CourseInstance.objects.get_primary_course_by_canvas_course_id')
+    @patch('mailgun.route_handlers.MailingList.objects.get_or_create_or_delete_mailing_list_by_address')
+    def test_multi_mailing_list_recipients_with_always_mail_staff_false(self, mock_ml_get, mock_ci_get, mock_ss_filter):
```
elliottyates: Ditto for docstring

mailgun/tests.py
```diff
@@ -269,20 +269,156 @@ def test_empty_body_html(self, mock_ml_get, mock_ci_get, mock_ss_filter):
...
+            original_cc_address=[],
+            original_to_address=[u'class-list@example.edu', u'bogus@example.edu'],
+            text=u'blah blah'
+        )
+        ml.send_mail.assert_has_calls([send_mail_call, send_mail_call])
+
+    @patch('mailgun.route_handlers.SuperSender.objects.filter')
+    @patch('mailgun.route_handlers.CourseInstance.objects.get_primary_course_by_canvas_course_id')
+    @patch('mailgun.route_handlers.MailingList.objects.get_or_create_or_delete_mailing_list_by_address')
+    def test_multi_mailing_list_recipients_with_always_mail_staff_true(self, mock_ml_get, mock_ci_get, mock_ss_filter):
```
elliottyates: Ditto for docstring

mailgun/tests.py
```diff
@@ -315,17 +451,17 @@ def test_multi_mailing_list_recipients(self, mock_ml_get, mock_ci_get, mock_ss_f
         response = handle_mailing_list_email_route(request)
         self.assertEqual(response.status_code, 200)
         send_mail_call = call(
-            'Unit Test via Canvas',
-            'unittest@example.edu',
-            ['teacher@example.edu', 'student@example.edu', 'unittest@example.edu'],
-            '[Lorem For Beginners] blah',
+            u'Unit Test via Canvas',
```
elliottyates: Why did these all change to unicode, but not the others? Just curious.

mailing_list/migrations/0013_auto_20160718_1945.py
```diff
@@ -0,0 +1,34 @@
...
+    dependencies = [
+        ('mailing_list', '0012_supersender'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='CourseSettings',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('canvas_course_id', models.IntegerField()),
```
elliottyates: Can we ditch the `id` field and just use this as the `primary_key`? Would we ever store multiple CourseSettings records for a single `canvas_course_id`? (Maybe only if we were versioning/saving all past versions of the settings...?)

mailing_list/migrations/0013_auto_20160718_1945.py
```diff
@@ -0,0 +1,34 @@
...
+                ('always_mail_staff', models.NullBooleanField(default=True)),
```
elliottyates: Any reason this is a `NullBooleanField` instead of a `Boolean` field?

mailing_list/models.py
```diff
@@ -168,6 +168,17 @@ def get_or_create_or_delete_mailing_lists_for_canvas_course_id(self, canvas_cour
         return result
 
 
+class CourseSettings(models.Model):
+    canvas_course_id = models.IntegerField()
+    always_mail_staff = models.NullBooleanField(null=True, default=True)
+    modified_by = models.CharField(null=True, max_length=32)
+    date_created = models.DateTimeField(null=True, default=timezone.now)
```
elliottyates: We could also use `auto_now_add=True` if we don't ever expect a user to be manually setting the `date_created` field.

mailing_list/models.py
```diff
@@ -168,6 +168,17 @@ def get_or_create_or_delete_mailing_lists_for_canvas_course_id(self, canvas_cour
...
+    date_modified = models.DateTimeField(null=True, default=timezone.now)
```
elliottyates: I think we want `auto_now=True` instead of `default`, right? Otherwise this will only be automatically updated on create (not on updates), I think.

mailing_list/migrations/0013_auto_20160718_1945.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
```
elliottyates: Can we change the migration file name to be more descriptive?

mailgun/route_handlers.py
eparker71: yes, good catch

mailgun/tests.py
eparker71: the tests were failing and the only diff I could see was the actual call and expected calls were diff in that one had u'' and one did not. It turns out there was another diff, the order of the lists were diff, but I didn't see that at the time. The tests pass now, so I didn't bother to change the strings back.

mailing_list/migrations/0013_auto_20160718_1945.py
eparker71: This is the default migration created when you run makemigrations. May we could do some XP on this? I'm not as familiar with the generated migrations code?

elliottyates: Sure. I think it's fine to leave it the way it is and investigate a bit for future knowledge as we have time during ACP etc.

mailgun/tests.py
eparker71: good catch, doc string updated

mailgun/tests.py
eparker71: ditto doc string

mailgun/tests.py
eparker71: ditto doc string

mailing_list/migrations/0013_auto_20160718_1945.py
eparker71: at the time there was, but I don't remember now :) I can convert it to a Boolean.

mailing_list/migrations/0013_auto_20160718_1945.py
eparker71: yeah, I renamed it to 0013_add_course_settings_model.py

mailing_list/migrations/0013_auto_20160718_1945.py
eparker71: ah, yeah, got it, that makes sense.

mailing_list/models.py
```diff
@@ -179,11 +190,12 @@ class MailingList(models.Model):
     ACCESS_LEVEL_STAFF = 'staff'
 
     canvas_course_id = models.IntegerField()
+    course_settings = models.ForeignKey(CourseSettings, null=True)
     section_id = models.IntegerField(null=True)
     access_level = models.CharField(max_length=32, default='members')
     created_by = models.CharField(max_length=32)
     modified_by = models.CharField(max_length=32)
-    date_created = models.DateTimeField(blank=True, default=timezone.now)
+    date_created = models.DateTimeField(auto_now_add=True)
```
elliottyates: This looks safe -- I don't see anywhere that it is overridden. We should probably update `date_modified` here and in the `CourseSettings` model to be consistent.

elliottyates: (we will also have to remove the modified date update line in `api.set_access_level()`)

mailing_list/migrations/0013_auto_20160719_1611.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
```
elliottyates: File name still needs updating (will do it in my commits)

PR #170, opened 2016-11-15
TLT-2802: route handler - fallback on from address if sender is not a member

elliottyates:
- for all lists but readonly lists, if sender is not a member of the list, but from address is a member and sender is an active email communication channel for the member in Canvas, forward on the message and use the from address as the new sender
- if sender is not an active email communication channel in Canvas, provide a specific, helpful bounceback email with instructions on how to setup this type of delegation
- if from address is a super sender and does not match sender address but sender address is an active email communication channel for the member in Canvas with a primary email matching the from address, forward on the message and use the from address as the new sender
- requires https://github.com/penzance/canvas_python_sdk/pull/74

https://jira.huit.harvard.edu/browse/TLT-2802

@cmurtaugh @sapnamysore


mailgun/route_handlers.py
```diff
@@ -180,32 +180,66 @@ def _handle_recipient(request, recipient):
...
-            bounce_back_email_template = 'mailgun/email/bounce_back_readonly_list.html'
-        elif ml.access_level == MailingList.ACCESS_LEVEL_STAFF and sender_address not in teaching_staff_addresses:
-            logger.info(
-                u'Sending mailing list bounce back email to %s for mailing list %s '
-                u'because the sender was not a staff member', sender, recipient)
-            bounce_back_email_template = 'mailgun/email/bounce_back_access_denied.html'
+    # if sender is a super sender or the list permission is open to everyone,
+    # proceed without checking for list membership
+    if sender_address not in super_senders \
+            and ml.access_level != MailingList.ACCESS_LEVEL_EVERYONE:
```
elliottyates: Should we also be sending as 'from' in the case of `ml.access_level == MailingList.ACCESS_LEVEL_EVERYONE` (regardless of whether the sender or from address are list members)?

mailgun/route_handlers.py
```diff
@@ -180,32 +180,66 @@ def _handle_recipient(request, recipient):
...
-                u'because the list is readonly', sender, recipient)
-            bounce_back_email_template = 'mailgun/email/bounce_back_readonly_list.html'
-        elif ml.access_level == MailingList.ACCESS_LEVEL_STAFF and sender_address not in teaching_staff_addresses:
-            logger.info(
-                u'Sending mailing list bounce back email to %s for mailing list %s '
-                u'because the sender was not a staff member', sender, recipient)
-            bounce_back_email_template = 'mailgun/email/bounce_back_access_denied.html'
+    # if sender is a super sender or the list permission is open to everyone,
+    # proceed without checking for list membership
+    if sender_address not in super_senders \
```
sapnamysore: The current changes don't apply if you are a  super sender. But should it? What if you are sending it from a different account(like gmail) configured  with an alias to the harvard account , and it's the harvard account that is in the super sender list?

elliottyates: Good question -- I think that super senders are a very select group of folks, and they probably don't need this feature. I will confirm with the PO!

mailgun/route_handlers.py
elliottyates: PO confirms this would be good to have as part of the story.

mailgun/route_handlers.py
```diff
@@ -314,22 +351,26 @@ def _get_attachments_inlines(request):
...
+    # enrollment; note even this won't work for anyone not enrolled in the
+    # course (ie. supersenders, globally sendable lists).
     if not sender_display_name:
-        name = get_name_for_email(ml.canvas_course_id, parsed_sender.address)
-        if name:
-            sender_display_name = name
-            logger.debug(u'Looked up display name for sender: %s, found: %s',
-                         parsed_sender.address, sender_display_name)
+        sender_display_name = get_name_for_email(ml.canvas_course_id,
+                                                 reply_to_address)
```
sapnamysore: Should this be reply_to_address.address?

mailgun/route_handlers.py
cmurtaugh: Seems reasonable!

mailgun/route_handlers.py
```diff
@@ -314,22 +351,26 @@ def _get_attachments_inlines(request):
...
+        if sender_display_name:
+            logger.debug(
+                u'Looked up enrollment information to find display name for '
+                u'sender %s, found: %s', reply_to_address, sender_display_name)
```
sapnamysore: Here as well? reply_to_address.address?

elliottyates:
The [latest commit, `be9a226`](https://github.com/Harvard-University-iCommons/lti_emailer/pull/170/commits/be9a226d80876412c5601561833a8fe5d046c038), adds appropriate from/sender handling for super senders and list members of public lists as well, per feedback in the [story conversation](https://jira.huit.harvard.edu/browse/TLT-2802) and per decision of PO and tech lead.

Now requires https://github.com/penzance/canvas_python_sdk/pull/74.

Description of new behavior and requirements has been added to the PR description.

sapnamysore:
👍

PR #172, opened 2016-12-01
tlt-2802: route handler - CommChannelCache fix

elliottyates:
- CommChannelCache was caching API call results for the lifetime of the app; with this fix, it only caches for the lifetime of a single handle_mailing_list_email_route() request
- minor refactoring of from fallback checks; moving low-cost comparison ahead of less-low-cost membership check

The CommChannelCache fix is [just a few lines at the bottom of this diff](https://github.com/Harvard-University-iCommons/lti_emailer/compare/task/elliottyates/tlt-2802/from_address_fallback_caching_fix?expand=1#diff-d829a7ff5d8e8e6b258e70e9a7da8c9fL414). The rest is the refactoring.

### DIFF
--- a/mailgun/route_handlers.py
+++ b/mailgun/route_handlers.py
@@ -194,8 +194,8 @@ def _handle_recipient(request, recipient, user_alt_email_cache):
     # any validation that fails will set the bounce template
     bounce_back_email_template = None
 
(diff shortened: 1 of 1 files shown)


sapnamysore:
👍

PR #173, opened 2017-01-24
tlt-2944: account delegation fix for users with multiple canvas accounts

elliottyates:
- when checking for valid delegate email address senders (envelope sender), consider all active email communication channels for canvas users with primary email address equal to 'from' as valid senders (formerly we denied sending when we found multiple canvas user accounts with the same primary email address)


sapnamysore:
Looks good! 👍

PR #176, opened 2017-01-24
tlt-2959: tooltip placement fix for member list link

elliottyates:
Admin page course and section member list links now show hover tooltip on right-hand side instead of above, to avoid an issue where the left edge of the iframe (in Canvas) was cutting off the tooltip.

An example of what it looks like with this fix:
screen shot 2017-01-24 at 4 26 20 pm


hktest:
:+1:  This view is much better than alternative (cut off text). Will touch base with @Vittorio2015 to make sure there are no concerns with the right placement.  Thanks much!

hktest:
Reviewed and OKed by @Vittorio2015 - thanks!

PR #181, opened 2018-05-31
Task/sapnamysore/tlt 3370/fix error msg

sapnamysore:
This PR is pulling in the change in icommons-ui  to the current 'Unauthorized' message as per AC #3 in https://jira.huit.harvard.edu/browse/TLT-3370 

Lti_emailer  is using this branch and is deployed on dev and the fix can be tested in this course https://canvas.dev.tlt.harvard.edu/courses/5809/external_tools/143  logged in as test12@mcelroy.org

Note that the qa version of the tool still has the old message


cmsnowden:
looks good

PR #182, opened 2018-08-08
Update troubleshooting urls

Chris-Thornton-Harvard:
Deployed and tested on dev


Chris-Thornton-Harvard:
Tested on dev and got Carters blessing

PR #184, opened 2018-10-29
Feature/handle batv senders

cmurtaugh:
This PR addresses TLT-3509: some email senders use "Bounce Address Tag Validation", which means that the address in the envelope sender field has a prefix added that looks like "prvs=<hash>=". 
This PR makes the Emailer work with these addresses by stripping off the prefix before doing any comparison with the list members, etc. 
The PR also includes a script that I used to test to make sure that the change works as expected. 
Finally -- I also updated the version of `pyopenssl` to address a security vulnerability (CVE-2018-1000807).

### DIFF
--- a/batv_test.py
+++ b/batv_test.py
@@ -1,0 +1,23 @@
+# Import smtplib for the actual sending function
+import smtplib
+
+# Import the email modules we'll need
+from email.mime.text import MIMEText
+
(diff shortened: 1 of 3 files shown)


sapnamysore:
Looks good 👍 Validated using the script and I am seeing the emails delivered correctly and dont see any more exception in splunk logs

PR #186, opened 2019-10-28
Hotfix/v1.18.1

cmurtaugh:
Update `Flanker` and unpin `cryptography` to get this to install on Ubuntu 18.04. 

Also switch to SSM Parameter Store for settings.


sapnamysore:
👍

Chris-Thornton-Harvard:
Looks good. Just to clarify, flanker does not require cryptography anymore?

cmurtaugh:
Flanker does require cryptography, but we can let it get installed via flanker’s requirements rather than pinning it to a specific version ourselves. 

That will result in (currently) cryptography 2.8 being installed, and that should work both on ubuntu 16.04 and 18.04.

PR #188, opened 2019-10-28
Updates to work with the latest flanker

cmurtaugh:
PR includes:
* Fix for setting the display_name (can't set it directly anymore; need to include the display name in the string to parse)
* `import address as addresslib_address` to avoid collisions


Chris-Thornton-Harvard:
👍

PR #200, opened 2021-07-27
TLT-3994: Remove option for world-access

PatrickMTonne:
Per [TLT-3994](https://jira.huit.harvard.edu/browse/TLT-3994), this PR removes the "World Access" option from LTI Emailer. 

This branch has been deployed to both Dev and QA. For an example of this update, see [this course](https://canvas.dev.tlt.harvard.edu/courses/171/external_tools/3) and click on the gear icon on the right-hand side of the "Email Entire Course" section. You will no longer see the "World Access" option within the list.

### DIFF
--- a/mailing_list/api.py
+++ b/mailing_list/api.py
@@ -6,6 +6,7 @@ from django.views.decorators.http import require_http_methods
 from django.contrib.auth.decorators import login_required
 from django.utils import timezone
 from django.core.cache import cache
+from django.core.exceptions import PermissionDenied
 
 from django_auth_lti.decorators import lti_role_required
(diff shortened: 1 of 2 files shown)


elliottyates:
👍 I'm thinking this is good enough, as it still supports courses that already have this setting enabled. An additional paranoid step would be shutting off the backend capability to do this for new calls to the API as well, e.g. explicitly validation in the [mailing_list.api.set_access_level()](https://github.com/Harvard-University-iCommons/lti_emailer/blob/v2.1.1/mailing_list/api.py#L57) handler.

PR #201, opened 2021-12-14
Fix recipient header

cmurtaugh:
Mailgun recently (Nov. 2021?) released a reimplemented version of their API which behaves slightly differently than the old one did. The impact to us is that messages sent through the emailer are received with a `To` header that contains the string `%recipient%` rather than the actual recipient address. 

Mailgun support suggested a change to the way that we set the `To` header in order to get the old behavior back. 

Essentially the change is to use Mailgun's `user-variables` feature to populate the `To` header with the mailing-list address. Mailgun will always add the actual recipient's individual email address, so we end up with the desired header: `users-address@harvard.edu, list-address@coursemail.harvard.edu`.

### DIFF
--- a/lti_emailer/requirements/base.txt
+++ b/lti_emailer/requirements/base.txt
@@ -1,4 +1,4 @@
-Django==2.2.4
+Django==2.2.24
 django-angular==2.2.1
 django-cached-authentication-middleware==0.2.2
 django-redis-cache==2.0.0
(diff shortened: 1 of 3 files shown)


PatrickMTonne (approved):
👍  Looks good!

elliottyates (approved):
👍🏻 thanks for the useful PR comments!

Chris-Thornton-Harvard:
👍

PR #207, opened 2022-07-25
Task tlt 4215 haydn ady lti emailer member count incorrect in dev

haydn9000:
* Added alert in list details page (after clicking on view members) about how not all the members in this list will be emailed. Only the ones in the whitelist in the database.

* Added comment in models.py members() function (quick note on why the condition in there).

Note: This change has been deployed in dev and qa Canvas. Here are links to [qa](https://canvas.qa.tlt.harvard.edu/courses/90925/external_tools/68) and [dev](https://canvas.dev.tlt.harvard.edu/courses/39652/external_tools/3) to test courses where this change has been tested.

Tested locally by temporarily modifying the code to include hardcoded URL to mock prod to see if message would disappear as expected. 
Temporary modified code used in app.js to test:
```
ml.isNonProdURL = function() {
      var URL = "https://lti-emailer.tlt.harvard.edu:8443/mailing_list/admin_index/";
      return URL.indexOf("lti-emailer.tlt") >= 0 ? false : true;
    };
```


Chris-Thornton-Harvard:
👍  Looks good!

PR #209, opened 2022-08-04
Remove member counts; move ML address to CC field

cmurtaugh:
This PR removes the list member counts from the main admin view. The link text is changed from `N members` to `view members`; the link goes to the member list view as usual. 

This PR also makes a change to outgoing messages processed by the tool: the mailing-list address is added to the `CC` header instead of to the `To` header as before. The prior behavior resulted in the message having two `To` headers, and Gmail has started to balk at this. Moving the mailing-list address to the `CC` field should be functionally equivalent for users. 

Finally, this PR contains some changes to the `log_post_data` webhook handler so that it will work with both legacy and current Mailgun webhooks. 

Closes TLT-4288 and TLT-4289.


Chris-Thornton-Harvard:
Tested on a [large course (1200+)](https://canvas.qa.tlt.harvard.edu/courses/6360) in QA and it loaded immediately.
Clicking into "view members did result in a long load time and eventual 502. A subsequent attempt to clicking into that worked fine due to caching.

cmurtaugh:
Yeah - I'd like to figure out how to speed up the member-list view too but we can tackle that separately. I'm also a little worried that the slowness of loading the member list might also be impacting the Mailgun route-handler; we must be fetching that list when we receive the API call from Mailgun, and it seems likely that some of those calls might be timing out too. Fortunately MG will retry those calls, so hopefully the data is cached by the time the retry happens.

Chris-Thornton-Harvard:
👍

PR #217, opened 2023-02-03
TLT-4382: Update to use lti_school_permissions

PatrickMTonne:
This PR:

- Adds the `django-canvas-lti-school-permissions` dependency
- Replaces references to `lti_permissions` (from `django_icommons_common`) with the library above.

The permission records from the old table were migrated into the new one via a [migration script](https://github.com/Harvard-University-iCommons/django-canvas-lti-school-permissions/blob/a1aa942ff04f3663f58f492539004100b8079e12/lti_school_permissions/migrations/0003_school_permission_custom_data.py) in `lti_permissions`. This script takes into account the slight differences between the two tables (`lti_permissions` has an `allow` field and accepts wildcards (`*`) in the `school_id` field; `lti-school-permissions` supports neither).

The script translates records with `*` to a record for each school [here](https://github.com/Harvard-University-iCommons/django-canvas-lti-school-permissions/blob/a1aa942ff04f3663f58f492539004100b8079e12/lti_school_permissions/migrations/0003_school_permission_custom_data.py#L69-L78).

Deployed to dev (example test course [here](https://harvard.beta.instructure.com/courses/96827/external_tools/23649))


mailing_list/migrations/0015_alter_mailinglist_access_level.py
Chris-Thornton-Harvard, line 1: What is the background on this migration?

PatrickMTonne: We can actually close this PR since this story was later designated as `Won't Do` (for reasons that escape me at the moment).

I don't recall exactly where it came from, but I believe I ran `makemigrations` for the `lti_school_permissions` project (or some migrations-related command) for the purposes of this story, and it resulted in a new migration for this project as well.

cmurtaugh:
We discussed this on 8.1.23 and decided that we do want to go ahead with this change (remove `django-icommons-common` references and use the `django-canvas-lti-school-permissions` library instead. 
We'll need to convert the permissions data from the old format to the new but can do that as a one-off script rather than via migrations.

Chris-Thornton-Harvard:
Created [TLT-4526](https://jira.huit.harvard.edu/browse/TLT-4526) to track DB permission migrations

PR #227, opened 2023-11-02
Task/haydn9000/tlt 4556/management cmd for report of use

haydn9000:
Django management command designed to create a CSV report of courses that have utilized the Course Emailer tool within a specified school and academic years.


Chris-Thornton-Harvard:
PR looks good, but I would hold on merging until we hear back from the user on what they would find useful in the report

PR #233, opened 2026-04-22
Add AWS_S3_CUSTOM_DOMAIN setting

cmurtaugh:
Add AWS_S3_CUSTOM_DOMAIN setting to `aws.py` so that staticfiles requests will go through CloudFront instead of directly to the bucket.


Chris-Thornton-Harvard:
👍

PR #234, opened 2026-05-05
v1 -> v2 `MailingList` table migration

PatrickMTonne:
## Context

Adds a one-time Django management command, `copy_to_dynamodb`, that migrates v1 `MailingList` records from PostgreSQL into the v2 Course Emailer's DynamoDB table so existing mailing lists continue to work after the v1 → v2 cutover. The command copies every `MailingList` row modified within a configurable window (default: the last 2 years).

## Summary

- New command `mailing_list/management/commands/copy_to_dynamodb.py` with flags:
  - `--table-name` (required) — target DynamoDB table
  - `--dry-run` — skip `put_item` calls
  - `--overwrite` — replace existing items instead of skipping
  - `--max-age-years N` — only copy lists modified within N years (default 2; 0 disables)
  - `--output-file PATH` — per-list CSV audit trail
- Idempotent by default via `ConditionExpression="attribute_not_exists(PK)"`; existing v2 items are preserved and counted as `skipped_existing`
- Non-zero exit code when any errors are recorded
- Field mapping from v1 to v2:
  - `access_level=members`/`everyone` → `valid_senders=all_members`, `active=True`
  - `access_level=staff` → `valid_senders=staff_only`, `active=True`
  - `access_level=readonly` → `valid_senders=all_members`, `active=False`
  - `course_settings.always_mail_staff` → `include_all_staff` (defaults to `True` when no `CourseSettings` row exists)
  - `date_created`/`date_modified` → ISO 8601 UTC `created_at`/`updated_at`
  - `section_id` included only when present; `subject_prefix`, `sis_section_id`, and `name` are omitted (v2's own `sync_with_canvas` backfills them on first course access and won't clobber migrated settings)
- `boto3` added as a direct dependency in `pyproject.toml`

## Testing

1. Dry-run against the v2 dev DynamoDB table with a CSV audit trail: `ENV=dev DJANGO_SETTINGS_MODULE=lti_emailer.settings.local uv run python manage.py copy_to_dynamodb --table-name <dev-table> --dry-run --output-file dry-run.csv`
2. Real run against dev without `--dry-run`; inspect the DynamoDB table and confirm new items have correct `PK`/`SK`, `active`, `valid_senders`, `include_all_staff`, ISO UTC timestamps, and no `subject_prefix`/`sis_section_id`/`name`
3. Load a migrated course in the v2 frontend and confirm the mailing list appears with correct access-level settings.

## Notes

- v1 PROD shows 16,217 modified email lists from the past two years:
```sql
  SELECT COUNT(*) FROM ml_mailing_list
  WHERE date_modified >= NOW() - INTERVAL '2 years';
```

### DIFF
--- a/mailing_list/management/commands/copy_to_dynamodb.py
+++ b/mailing_list/management/commands/copy_to_dynamodb.py
@@ -16,7 +16,10 @@ ACCESS_LEVEL_TO_V2 = {
     MailingList.ACCESS_LEVEL_MEMBERS: {"valid_senders": "all_members", "active": True},
     MailingList.ACCESS_LEVEL_EVERYONE: {"valid_senders": "all_members", "active": True},
     MailingList.ACCESS_LEVEL_STAFF: {"valid_senders": "staff_only", "active": True},
-    MailingList.ACCESS_LEVEL_READONLY: {"valid_senders": "all_members", "active": False},
+    MailingList.ACCESS_LEVEL_READONLY: {
+        "valid_senders": "all_members",
+        "active": False,
+    },
 }
 
 ACTION_WRITTEN = "written"
--- a/mailing_list/management/commands/copy_to_dynamodb.py
+++ b/mailing_list/management/commands/copy_to_dynamodb.py
@@ -1,0 +1,206 @@
+import csv
+import logging
+import sys
+from datetime import timezone as dt_timezone
+
+import boto3
+from botocore.exceptions import ClientError
+from django.conf import settings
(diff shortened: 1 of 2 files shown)


andresmauro17 (approved):
I was testing this locally using the CSV output file and the DynamoDB table.

All 31 existing records were copied correctly.

It should be fine if we can test all the records in the DB, and the code looks like it will work as expected.

Good implementation @PatrickMTonne  👍

